PR #31, opened 2023-10-08
Fix Album Sorting for multiple Discs

PalanixYT:
When an album has multiple discs, then it will not sort the items correctly, 
because there will be multiple songs with the same track.

This patch tries to alleviate that, without adding any new logic


nift4:
thanks

PR #34, opened 2023-10-22
Changed collapsed toolbar text color

HotarunIchijou:


AkaneTan:
@HotarunIchijou @nift4 can you compare two color tokens and make a decision?

nift4:
good catch, thanks

PR #72, opened 2024-01-02
Added Nepalese translation

ankitstha:


AkaneTan:
Please submit via our weblate. Translation via PR is not encouraged.

PR #76, opened 2024-01-13
fix: convert Uri to File

bggRGjQaUbCoE:
In some cases, Uri goes like "/storage/emulated/0/Music/*", but toFile requires the scheme is file


nift4:
How to reproduce the bug?

bggRGjQaUbCoE:
> How to reproduce the bug?

It appears in OneUI 5

PR #86, opened 2024-02-16
Support for repeated lyrics on LRC format

strongville:
Some programs, such as [foobar2000's OpenLyrics plugin](https://github.com/jacquesh/foo_openlyrics), have the ability to save "compressed" LRC files, making that if a line on the song is repeated exactly, instead of being added in sequence for the song, a time marker is added to the same previous line on the song.

image

Previous to this change, for this kind of lines, the lyrics viewer for Gramophone showed it only once, the first time that appeared, along with the secondary markers on display. This makes sure that all lines are now rendered properly.

(Little benchmark with this change, average 5 times same file: before 34ms, after 44ms)


nift4:
thanks

PR #123, opened 2024-02-29
Fixes #110

### LINKED ISSUE #110: [Milestone] Fix inconsistent space on both sides of grid layout recyclerview
Should be 24 dp, but now it's something like 12

VishnuSanal:
Fixes #110


AkaneTan:
https://github.com/AkaneTan/Gramophone/blob/ccb1b1d62e419ec4f397fc30bcb1bd6cdf507f8b/app/src/main/kotlin/org/akanework/gramophone/ui/components/GridPaddingDecoration.kt#L13
This is what changes the side space of grid layout, simply adding margin won't do the trick

PR #135, opened 2024-03-14
Support multiline LRC

mikooomich:
Hi. Usually LRC files expect a sync timestamp on each line, however in some cases it may be better to relax this strict syntax (i.e. self created LRCs, formatting, etc.), see below for example. This setting preference is **disabled** by default. When enabled, this should not change how non multiline lines are displayed.

Sample lyrics:
```
[00:11.22] hewwo i am lyric
hewwo i am second line
[00:22.33] hewwo i am the next block

[00:33:44] hewwo i am a separate section so the previous
lyric has extra \n spacing displayed
```

Should display as:
```
hewwo i am lyric
hewwo i am second line
hewwo i am the next block

hewwo i am a separate section so the previous
lyric has extra \n spacing displayed
```


strongville:
Honestly, I haven't heard of any lyrics software that does this when having several lines that should be displayed at the same time. Sounds like a misbehave of the software (or person) doing this.

For that matter, putting same timestamp for several lines should work just fine for Gramophone, and also other software, like OpenLyrics, Poweramp, etc.

mikooomich:
> Honestly, I haven't heard of any lyrics software that does this ...

I've considered this, hence the opt in and and preserving single line aspect. It's not in any guidelines, but (to my knowledge) won't change non-multiline either, why not have it as an option? [Musicolet ](https://krosbits.in/musicolet/)does partially support multiline though, as in all the lines without timestamps get displayed, but as separate, sync-less lines in between sync points. Their old style parses akin to this PR.

Putting the same timestamp does not work properly in terms of highlighting on Gramophone. Having the same timestamp for multiple lyrics can have weird behaviors depending on implementation, ex reading the last one only

strongville:
> [Musicolet ](https://krosbits.in/musicolet/)does partially support multiline though,

From what I've seen trying Musicolet, it seems that the multiline parsing you're making reference to, it's just an incomplete file made by the built-in lyrics editor. Just saying that maybe it's not needed to support this kind of lyrics outside that.

> can have weird behaviors depending on implementation

I can speak for Gramophone's implementation regarding that, all consecutive timestamps are treated as translations, according to the UI already present on the app.

Issue #154, opened 2024-04-08
 [FR] Rearrange Titles in Queue Using Gestures


TechD123:
**Is your feature request related to a problem? Please describe.**
Rearranging the order of the queue is not possible in an easy manner. There is the possibility to use the "play next" feature for this, but it isn't nearly as efficient as using gestures.

**Describe the solution you'd like**
Quite often you want to change the order of the songs in the queue, which is not possible at the moment. One way to elegantly solve this would be by having a way to tap and drag a song to the desired position. For an example, take a look at Retro/Metro music player.

**Describe alternatives you've considered**
Gestures are likely to be the most intuitive solution.

Thanks for the great app!

nift4:
@TechD123 Yep this is planned, thanks for your request.

TechD123:
Awesome! Loving the clean design of the app by the way.

nift4:
This has been added to the code and will be available next release

PR #183, opened 2024-05-25
Rearrange Titles in Queue Using Gestures

topazrn:
Implement #154


nift4:
Hi, thanks for this PR - not updating the unshuffled index will cause playlist to be out of sync like in the bug you reported, but i'll commit a fix

topazrn:
Oops, I didn't notice there was such a thing. Thanks for the incoming fix!

nift4:
It turned out to be a little more complicated, I'll commit a fix some other day

nift4:
This lead me to discover some bugs in our shuffle implementation, thanks for that :P

PR #240, opened 2024-08-12
Update strings.xml

HanzoDev1375:
Hello, I am an Iranian and I translated the Persian translation according to the real Persian


nift4:
Thanks. next time you could also use https://hosted.weblate.org/engage/gramophone/

PR #314, opened 2024-10-31
Add shuffle all option in home menu - closes #313

### LINKED ISSUE #313: [FR] Show play all/shuffle all buttons in every tab
The play all and shuffle all buttons are displayed only in the "Songs" tab.

https://github.com/AkaneTan/Gramophone/blob/4d4d711ccd9ff2d1262ff70fe621052dd8a0db7c/app/src/main/kotlin/org/akanework/gramophone/ui/adapters/BaseDecorAdapter.kt#L62-L65
https://github.com/AkaneTan/Gramophone/blob/4d4d711ccd9ff2d1262ff70fe621052dd8a0db7c/app/src/main/kotlin/org/akanework/gramophone/ui/adapters/BaseDecorAdapter.kt#L174-L186

I was wondering if there were downsides to showing these buttons in other tabs as well.

lucaxvi:
- Adds a "shuffle all" option to the three dot menu.
- Fixes a bug with the shuffle all button in the songs fragment skipping some songs.

### DIFF
--- a/app/src/main/kotlin/org/akanework/gramophone/ui/adapters/BaseDecorAdapter.kt
+++ b/app/src/main/kotlin/org/akanework/gramophone/ui/adapters/BaseDecorAdapter.kt
@@ -34,7 +34,6 @@ import org.akanework.gramophone.R
 import org.akanework.gramophone.logic.ui.ItemHeightHelper
 import org.akanework.gramophone.logic.ui.MyRecyclerView
 import org.akanework.gramophone.ui.getAdapterType
-import kotlin.random.Random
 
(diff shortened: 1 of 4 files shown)


nift4:
Won't shuffle all now always start at the same song?

lucaxvi:
It looks like exoplayer is smart enough to start the playlist with a random song when shuffle mode is enabled and you call setMediaItems without passing a start index. I just tested it and it works as intended.

nift4:
That code does predate the introduction of CircularShuffleOrder by a few months so I guess that makes sense. Give me some time in general to get back to Gramophone and I'll probably merge this

PR #379, opened 2025-02-01
Add feature show bitrate in details fragment

tungnk123:


app/src/main/kotlin/org/akanework/gramophone/logic/GramophoneExtensions.kt
```diff
@@ -27,6 +27,7 @@ import android.content.res.Configuration
 import android.graphics.Color
 import android.graphics.drawable.AnimatedVectorDrawable
 import android.graphics.drawable.Drawable
+import android.media.MediaMetadataRetriever
```
nift4, line 30: Can you use https://developer.android.com/reference/androidx/media3/exoplayer/MetadataRetriever#retrieveMetadata(android.content.Context,androidx.media3.common.MediaItem) instead? ExoPlayer is more predictable as it's shipped as part of app code.

app/src/main/kotlin/org/akanework/gramophone/logic/GramophoneExtensions.kt
```diff
@@ -98,6 +99,25 @@ fun MediaItem.getFile(): File? {
     return getUri()?.toFile()
 }
 
+fun MediaItem.getBitrate(context: Context): Long {
```
nift4:
```suggestion
fun MediaItem.getBitrate(context: Context): Long? {
```

app/src/main/kotlin/org/akanework/gramophone/logic/GramophoneExtensions.kt
```diff
@@ -98,6 +99,25 @@ fun MediaItem.getFile(): File? {
...
+    val retriever = MediaMetadataRetriever()
+    return try {
+        val filePath = getFile()?.path
+        if (filePath != null) {
+            retriever.setDataSource(filePath)
+        } else {
+            val uri = requestMetadata.mediaUri ?: return 0
+            retriever.setDataSource(context, uri)
+        }
```
nift4: Is this else path really needed?

app/src/main/kotlin/org/akanework/gramophone/logic/GramophoneExtensions.kt
```diff
@@ -98,6 +99,25 @@ fun MediaItem.getFile(): File? {
...
+        val s = retriever.extractMetadata(MediaMetadataRetriever.METADATA_KEY_BITRATE)
+        s?.toLongOrNull() ?: 0
```
nift4:
```suggestion
        s?.toLongOrNull()
```

app/src/main/kotlin/org/akanework/gramophone/logic/GramophoneExtensions.kt
```diff
@@ -98,6 +99,25 @@ fun MediaItem.getFile(): File? {
...
+    } catch (e: Exception) {
+        0
```
nift4:
```suggestion
        Log.w("getBitrate", Log.getStackRateString(e))
        null
```

app/src/main/kotlin/org/akanework/gramophone/ui/fragments/DetailDialogFragment.kt
```diff
@@ -82,6 +84,8 @@ class DetailDialogFragment : BaseFragment(false) {
         mimeTypeTextView.text = mediaItem.localConfiguration?.mimeType ?: "(null)"
         pathTextView.text = mediaItem.getFile()?.path
             ?: mediaItem.requestMetadata.mediaUri?.toString() ?: "(null)"
+        val bitrate = mediaItem.getBitrate(requireContext())
+        bitRateTextView.text = if (bitrate > 0) "${bitrate / 1000} kbps" else "(no bitrate)"
```
nift4: Please make both the kbps string (Android supports format strings) and "(no bitrate)" translatable, and maybe consider saying "unknown" instead of "no"

nift4:
```suggestion
        bitRateTextView.text = if (bitrate != null) "${bitrate / 1000} kbps" else "(no bitrate)"
```
to allow distinguishing between 0kbps and really unknown value

nift4 (changes requested):
Thanks for the PR! Left a few comments

app/src/main/kotlin/org/akanework/gramophone/logic/GramophoneExtensions.kt
tungnk123: After researching, I found that `MetadataRetriever.retrieveMetadata()` does not provide bitrate information for local files; it primarily retrieves embedded metadata like ID3 tags. To obtain the actual bitrate, ExoPlayer must be running and playing the media. On the other hand, `MediaMetadataRetriever` works for local files without requiring playback.

app/src/main/kotlin/org/akanework/gramophone/logic/GramophoneExtensions.kt
tungnk123: Yes, the `else` path is needed to handle cases where the media does not have a local file path but instead has a valid `content://` URI, which is common for media retrieved from Android's MediaStore. This ensures that `MediaMetadataRetriever `can still extract metadata when the file path is unavailable.

tungnk123:
@nift4 Thanks very much for these informative reviews!

app/src/main/kotlin/org/akanework/gramophone/logic/GramophoneExtensions.kt
nift4: In general, yes, but in Gramophone's case _every_ media item should have a file path because we query the real path from MediaStore when querying songs. Many things in the app assume this, and it's fine to do so here as well. Unless it doesn't work without the else path, which would mean I missed something, which can happen. Hence: is this else _really_ needed? Or does it work without? 😉

tungnk123: Let me check it out

tungnk123: Great! Looks like it still works fine without the else path. I'll keep it simplified unless we run into any unexpected edge cases. Thanks for the review! 🚀

app/src/main/kotlin/org/akanework/gramophone/logic/GramophoneExtensions.kt
```diff
@@ -98,6 +100,21 @@ fun MediaItem.getFile(): File? {
...
 
+fun MediaItem.getBitrate(): Long? {
+    val retriever = MediaMetadataRetriever()
+    return try {
+        val filePath = getFile()?.path ?: return null
+        retriever.setDataSource(filePath)
+        retriever.extractMetadata(MediaMetadataRetriever.METADATA_KEY_BITRATE)
+            ?.toLongOrNull()
+    } catch (e: Exception) {
+        Log.w("getBitrate", e.printStackTrace().toString())
```
nift4:
```suggestion
        Log.w("getBitrate", Log.getStackTraceString(e))
```
Sorry for the typo earlier lol, after this can merge

tungnk123: No problem

nift4:
Thanks for your contribution!

PR #407, opened 2025-02-17
Hide null bit depth in player

SharnavM:
Some audio formats (like mp3 and m4a) uses lossy codecs so the bit depth is lost. Currently the player displays "nullbit", so we conditionally render the bit depth if the information is present.


nift4:
Bit depth isn't the only value which can have invalid data and I was planning to address all of them at once but I guess it doesn't harm to merge this now. Thanks for the PR!

Issue #560, opened 2025-06-26
[FR] Modify the contributor list to a card-style


ghhccghk:
I am trying to modify the contributor interface style to make it look better. The following picture shows the effect I have achieved so far.

![Image](https://github.com/user-attachments/assets/8ffd3bfb-b623-4555-9c91-dc9f37c1c075)

Currently only the username, avatar and bio are displayed

ghhccghk:
Adjustment completed

![Image](https://github.com/user-attachments/assets/767cae3d-1ecf-4825-bc4b-d97e45053caa)

PR #561, opened 2025-06-26
feat: Modernize contributors settings using Jetpack Compose

ghhccghk:
Contributor interface display #560 

Screenshot_2025-06-26-21-38-48-739_org akanework gramophone debug


nift4:
Hi, thank you very much for the PR. There's a small problem with the current approach: I did a poll in Gramophone user community and 25% of all users said internet permission is not acceptable.

Is it possible to keep the images in the repo and add a small python script to update them, which I can run manually?

ghhccghk:
> Hi, thank you very much for the PR. There's a small problem with the current approach: I did a poll in Gramophone user community and 25% of all users said internet permission is not acceptable.
> 
> Is it possible to keep the images in the repo and add a small python script to update them, which I can run manually?

I will try tomorrow to see if I can get information updates from the Internet while compiling

nift4:
There is also a problem when doing it while compiling, because F-Droid compiles the app without internet access (life is complicated :-). So it should be a script (Python, Bash, or Kotlin Script would be good) that just can be run whenever I want, and it adds the image files to repo.

ghhccghk:
> There is also a problem when doing it while compiling, because F-Droid compiles the app without internet access (life is complicated :-). So it should be a script (Python, Bash, or Kotlin Script would be good) that just can be run whenever I want, and it adds the image files to repo.

OK, let me see how to implement it using Python

ghhccghk:
Modification completed

nift4:
thanks, I will check the code soon (I'm a bit busy right now)

PR #614, opened 2025-07-27
update readme to use horizontal translation status badge

Fuan200:
Update README to use horizontal translation status badge.

### Before

image

### After

image


nift4:
Thanks for your PR, I'm not sure whether I prefer this one though. It seems quite overwhelming for first-time visitors?

Fuan200:
I understand, when I made the change, I was thinking it might help to motivate translators to contribute

nift4:
Hmm yes that might actually work

PR #615, opened 2025-07-30
Switch Media3 module import method

ghhccghk:
This commit enables the `androidxMediaEnableMidiModule` flag in `settings.gradle.kts` and switches the Media3 module import method.

Specifically:
- `settings.gradle.kts`:
- Set `gradle.extra[“androidxMediaEnableMidiModule”] = true`.
- Replace the `includeBuild` block for Media3 with `(gradle as ExtensionAware).extra[“androidxMediaModulePrefix”] = “media3-”` and `apply(from = file(“media3/core_settings.gradle”))`. This allows Media3 subprojects to be included in a more standard way.
- `app/build.gradle.kts`:
- Replace remote Media3 dependencies with local project dependencies (e.g., `implementation(project(“:media3-lib-exoplayer”))`).

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -2,7 +2,6 @@
 
 import com.android.build.api.dsl.ApplicationBuildType
 import com.android.build.gradle.tasks.PackageAndroidArtifact
-import org.gradle.kotlin.dsl.support.kotlinCompilerOptions
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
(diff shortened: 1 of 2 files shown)


nift4:
Hi, thanks for this PR, but using dependency substitution was actually on purpose as forked media3 is intended to be temporary (I'm always working on upstreaming patches), so I want to switch back to Google's media3 when all issues and features Gramophone needs have been addressed.
Dependency substitution allows me to just delete that block from one file and be back to latest Google media3.

PR #641, opened 2025-08-12
Improve landscape phone split screen ui

mikooomich:
Before
- Screenshot_20250812-133739_Gramophone
- Screenshot_20250812-133731_X-plore

After
- Screenshot_20250812-133443_Gramophone
- Screenshot_20250812-133502_Gramophone


nift4:
thats nice, does this also fix the case where portrait splitscreen with gramophone as top app uses landscape ui?

mikooomich:
Yep

app/src/main/java/org/akanework/gramophone/ui/components/FullBottomSheet.kt
```diff
@@ -1000,9 +1000,10 @@ class FullBottomSheet
                     ImageRequest.Builder(context).apply {
                         data(Pair(file, mediaItem?.mediaMetadata?.artworkUri))
                         if (bottomSheetFullCover.width == 0 || bottomSheetFullCover.height == 0) {
-                            throw IllegalStateException("expected to have box of full cover after layout")
+                            size(300, 300) // need some bitmap for material colour
```
nift4, line 1003: when did you reproduce a 0x0 sized image view?

the most common situation is during layout transitions (changing screen rotation or entering freeform or something idk), at which point no bitmap should be loaded yet.

loading a 300x300 bitmap just to start loading the correct sized bitmap a few frames later is something I wanted to avoid in the first place, which is why there is a throw there.

you said there's not enough space for the image view, in which configuration does that happen?

mikooomich: > you said there's not enough space for the image view, in which configuration does that happen?


Medium phone emulator, 50% or less split screen. The ui is too short in height in that case. You can also just lower the device's dpi ig

PR #655, opened 2025-08-21
Set the Format Information dialog box to Dynamic Color Picker.

ghhccghk:


nift4:
thanks for your pr, you inspired me to solve this using a less code-intensive way (using wrappedContext if its not null)

Issue #335, opened 2024-12-05
[FR] Hide tab bar when only one is present


ant14181:
**Is your feature request related to a problem? Please describe.**
I'd start with an example; you're an user whom only listen to music with shuffle on. At some point, you'd only want the "songs" tab to be the one present in the entire bar. Seeing an empty bar with only one tab strip seems redundant, wasting a tad bit of space that could be used for showing more songs.

**Describe the solution you'd like**
In case the user chooses only one tab to be shown, hide the tab bar completely; as proposed in the concept below (made with really little to no effort):
| Current | Proposed |
|--------|--------|
| ![image](https://github.com/user-attachments/assets/f3a23741-164c-4462-8aee-3f72965d4100) | ![image](https://github.com/user-attachments/assets/739636c4-a0e2-431a-a4d0-222ed70712a5)|

---
- [x] I will open one issue per request, follow the issue format and will add `[FR]` to the start of my issue title, and I acknowledge that if I don't follow the format, my issue might get closed without further explanation.

PR #728, opened 2025-10-11
[FR] #335

SurFace81:
Hides the tab bar when only one tab is present

### DIFF
--- a/app/src/main/java/org/akanework/gramophone/ui/fragments/ViewPagerFragment.kt
+++ b/app/src/main/java/org/akanework/gramophone/ui/fragments/ViewPagerFragment.kt
@@ -186,29 +186,31 @@ class ViewPagerFragment : BaseFragment(true) {
                 viewPager2
             )
         viewPager2.adapter = adapter
+
+        TabLayoutMediator(tabLayout, viewPager2) { tab, position ->
+            tab.text = getString(adapter.getLabelResId(position))
+            tab.view.post {
+                try {
+                    /*
+                     * Add margin to last and first tab.
+                     * There's no attribute to let you set margin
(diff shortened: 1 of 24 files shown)


app/src/main/java/org/akanework/gramophone/ui/fragments/ViewPagerFragment.kt
```diff
@@ -186,25 +186,30 @@ class ViewPagerFragment : BaseFragment(true) {
...
-                        resources.getDimension(R.dimen.tab_layout_content_padding).toInt() else 0
-                    lp.marginEnd = if (position == tabLayout.tabCount - 1)
-                        resources.getDimension(R.dimen.tab_layout_content_padding).toInt() else 0
-                    tab.view.layoutParams = lp
-                } catch (_: IllegalStateException) {
+        if (adapter.itemCount < 2) {
+            tabLayout.visibility = View.GONE
+        } else {
+            tabLayout.visibility = View.VISIBLE
+            TabLayoutMediator(tabLayout, viewPager2) { tab, position ->
```
nift4: Is there a specific reason you only attach the Mediator if there is atleast two tabs?
This causes following issue:
1. set preference to show only one tab
2. exit the app entirely
3. open the app, only one tab is shown so mediator isn't attached
4. now change preference to show two tabs

The tabs will not be rendered because Mediator is never attached. But moving it outside of if block and always attaching it seems to work fine

SurFace81: Yeah, it is an error

nift4:
Thanks for the PR!

PR #743, opened 2025-10-22
feat: Automatically fetches contributors from GitHub API

ghhccghk:
This commit modifies the avatar retrieval script by adding a function to automatically extract contributor lists from GitHub repository APIs.
- Added `fetch_contributors` function, which:
  - Fetches all repository contributors from GitHub API using pagination.
  - Filters out bot accounts and users already present in the manual list.
  - Appends new contributors to the `GITHUB_USERS` list for subsequent processing.
- Calls `fetch_contributors` within the `main` function to automatically update the list before generating the final contributors file.
- Adds new contributors and their avatar resources.


nift4:
Hi, thanks for the PR. 

There is a significant problem: GitHub API also shows translators as contributors. And there are too many translators to include every profile picture in the credits, the list will be extremely long and it will use a lot of storage space even with WebP images.
So it would be good if there was a way to get only people who changed code, but I think there is none.
So I downloaded credits from Weblate API instead, but there is also no way to understand which weblate user is which GitHub user (there is no requirement to use same username), so I can't use that to remove translators from that list either.

So I think manually maintaining credits is only solution, or do you have any other idea?

ghhccghk:
> Hi, thanks for the PR.
> 
> There is a significant problem: GitHub API also shows translators as contributors. And there are too many translators to include every profile picture in the credits, the list will be extremely long and it will use a lot of storage space even with WebP images. So it would be good if there was a way to get only people who changed code, but I think there is none. So I downloaded credits from Weblate API instead, but there is also no way to understand which weblate user is which GitHub user (there is no requirement to use same username), so I can't use that to remove translators from that list either.
> 
> So I think manually maintaining credits is only solution, or do you have any other idea?

There is currently no good solution.

nift4:
Okay, then I will close PR for now. But thank you for the attempt!

ghhccghk:
> Okay, then I will close PR for now. But thank you for the attempt!

However, I believe we can exclude cases where the username and nickname are identical.

PR #771, opened 2025-10-31
Add support for swipe-to-change-song on preview sheet

imjyotiraditya:


nift4:
weird, it seems to cause servere lags for me in a debug build (1.5k item playlist). with how many songs did you test it? sadly, profiler isn't working so I can't figure out why that happens at the moment.

nift4:
grafik
For some godforsaken reason my phone spends over 30s due to string comparisions. I've tested on another phone with almost the same songs (copied over 2k songs and only 100 songs difference) and I can't reproduce there, which makes the whole thing even weirder. Do you have any idea why that might possibly be?

nift4:
oh also,
- when entering an album page the text disappears
- with this PR you can no longer click on the text to open full player

any chance you can fix it? sorry for the long delay and thanks in advance!

PR #776, opened 2025-11-05
feat: Add playback speed control

ghost:
Adds a new button and dialog to the player which allows you to set a custom playback speed

### DIFF
--- a/app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
+++ b/app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
@@ -391,13 +391,13 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
                 .setPlaybackLooper(internalPlaybackThread.looper)
                 .build()
         )
-	    player.exoPlayer.addListener(object : Player.Listener {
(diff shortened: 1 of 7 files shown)


nift4:
Hi, thanks for the PR! Some thoughts from just looking at the diff:
1. I totally agree with locking speed and pitch as default, but similar to NewPipe there should probably be a second slider and a checkbox to unlock them. Being able to change them independently is convenient sometimes.
2. How did you arrive at maximum speed of 2x? I think audio buffer is allocated to support up to 8x speed increase if I remember correctly. Although you have to be quite crazy to listen at 8x so maybe UI does not need to go that high. Maybe 4x?
3. Button bar at the bottom is getting too cramped honestly. I already wanted to move sleep timer up there myself sometime. Can you move up speed control to the opposite side of the arrow button, at the top?
4. Speed and pitch need to be both saved and restored by LastPlayedManager. Reason is that it can be changed by external apps through MediaController API (although I don't know any example app, it makes sense to take care of edge cases) and playback can also be started without having activity open (playback resumption on Android 12 or later, or Bluetooth headset play button, for example). If saved and read in activity like this it will not behave consistenly depending on how you start playback which is confusing.

PR #810, opened 2025-12-31
feat: Added Music Island sharing support for Xiaomi's Hyper OS

ghhccghk:
A new utility class `IsLandHelp.kt` was introduced to build music sharing data for Xiaomi's Super Island.

In `GramophonePlaybackService`:
- When media metadata (song information) changes, `IsLandHelp.isLandMusicShare` is called to generate the corresponding `Bundle`.

In `MeiZuLyricsMediaNotificationProvider`:
- Modified the constructor to accept a `bundleProvider`.
- Added this `Bundle` to the notification's `extras` to enable Super Island functionality.


app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -599,6 +599,21 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
 	    return super.onStartCommand(intent, flags, startId)
     }
 
+    override fun onMediaMetadataChanged(metadata: MediaMetadata) {
```
nift4: inside MeiZuLyricsMediaNotificationProvider.kt, addNotificationActions, the parameter "session" has player and you can access with "session.player.mediaMetadata". Can you move Bundle generation there? Thanks!

ghhccghk: Let me see.

PR #865, opened 2026-03-15
Initial multiple queues support

mikooomich:
Adds "basic" support for multiple queues. Other needed features (ex. lastplayedmanger, queue menu, queue titles) will be added later, or much later in terms of nice to haves (ex. multiselect, search, etc).

To test: Enable in Flags.kt, then enable settings toggle

Screenshot_20260425-154337_Gramophone
Screenshot_20260425-154327_Gramophone
Screenshot_20260425-154348_Gramophone
Screenshot_20260425-154344_Gramophone


app/src/main/java/org/akanework/gramophone/logic/GramophoneExtensions.kt
```diff
@@ -337,6 +348,160 @@ fun MediaController.getAudioFormat(): AudioFormatDetector.AudioFormats =
...
+        }
+    }
+
+fun MediaController.getQueue(index: Int = C.INDEX_UNSET): MultiQueueObject? =
+    sendCustomCommand(
+        SessionCommand(SERVICE_QB_GET_QUEUE, Bundle.EMPTY).apply {
+            customExtras.putInt("index", index)
+        }, Bundle.EMPTY
+    ).get().extras.run {
+        if (containsKey("allQueues")) {
```
nift4: why bother with containsKey if you throw anyway, you can just let the !! do its job

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
```diff
@@ -0,0 +1,582 @@
...
+            Log.w(TAG, "Cannot find queue to delete: ${mq.title}")
+        }
+
+        return masterQueues.size
+    }
+
+    /**
+     * Deletes a queue.
+     *
+     * When deleting the active queue,
```
nift4: ,

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
```diff
@@ -0,0 +1,582 @@
...
+     *
+     * @param mq
+     */
+    fun deleteQueue(index: Int): Int {
+        if (QUEUE_DEBUG)
+            Log.d(TAG, "DELETING QUEUE AT INDEX: $index")
+        if (index == masterQueues.lastIndex) {
+            masterQueues.removeAt(index)
+            if (index <= 0) {
+                player.pauseAllPlayersAndStopSelf() // TODO: correct way to stop playback
```
nift4: just removeMediaItems()

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -156,6 +168,7 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
     val endedWorkaroundPlayer
         get() = mediaSession?.player as EndedWorkaroundPlayer?
     private var controller: MediaBrowser? = null
+    val qb: QueueBoard = QueueBoard(this)
```
nift4: maybe should be inited in onCreate

app/src/main/java/org/akanework/gramophone/logic/GramophoneExtensions.kt
```diff
@@ -337,6 +348,160 @@ fun MediaController.getAudioFormat(): AudioFormatDetector.AudioFormats =
...
+            customExtras.putInt("to", to)
+        }, Bundle.EMPTY
+    ).get().extras.run {
+        if (containsKey("status"))
+            getBoolean("status")
+        else throw IllegalArgumentException("expected status to be set")
+    }
+
+// TODO: shuffle and repeat mode
+fun MediaController.playQueue(
```
nift4: you are doing this at the wrong level, Android AUto for example will just not call your SERVICE_QB_ENQUEUE, it will keep doing setMediaItems(). Instead a player wrapper should give old queue to queueboard before executing setMediaItems()

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
```diff
@@ -0,0 +1,582 @@
...
+     * Retrieve the song at current position in the queue
+     */
+    fun getCurrentSong(): MediaItem? {
+        return queue.getOrNull(startIndex)
+    }
+
+    /**
+     * Retrieve a song given a song ID. Returns null if no song is found
+     */
+    fun findSong(mediaId: String): MediaItem? {
```
nift4, line 559: unused?

app/src/main/java/org/akanework/gramophone/ui/components/PlaylistQueueSheet.kt
```diff
@@ -75,30 +116,164 @@ class PlaylistQueueSheet(
...
+
+                                layout
+                            },
+                            update = { view ->
+                                val trigger = durationState.value
+                                val durationView: Chronometer = view.findViewById(R.id.duration)
+
+                                val current = (instance?.currentMediaItemIndex ?: 0)
+                                if (current > playlistAdapter.playlist.second.size) {
+                                    // hax to workaround ui loading itself 4 times
```
nift4: ?

app/src/main/java/org/akanework/gramophone/ui/adapters/BaseDecorAdapter.kt
```diff
@@ -187,20 +192,24 @@ open class BaseDecorAdapter<T : AdapterFragment.BaseInterface<*>>(
...
             } else if (adapter is AlbumAdapter) {
                 val list = adapter.getAlbumList()
                 val controller = adapter.getActivity().getPlayer()
                 controller?.repeatMode = REPEAT_MODE_OFF
                 controller?.shuffleModeEnabled = false
                 list.takeIf { it.isNotEmpty() }?.also { albums ->
-                    controller?.setMediaItems(albums.shuffled().flatMap { it.songList })
-                    controller?.prepare()
-                    controller?.play()
+                    controller?.playQueue(
```
nift4: cf, this ui code shouldnt be changed. isOriginal is always true after setMediaItems(), it should only be considered changed after add/remove/move(/replace if id changed)

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
```diff
@@ -0,0 +1,582 @@
...
+         */
+        val seamlessSupported = seamlessAllowed && (startIndex < mediaItems.size)
+                && plr.currentMediaItem?.mediaId == mediaItems[startIndex].mediaId
+        if (seamlessSupported) {
+            Log.d(TAG, "Trying seamless queue switch. Is first song?: ${startIndex == 0}")
+            val playerIndex = plr.currentMediaItemIndex
+
+            if (startIndex == 0) {
+                val playerItemCount = plr.mediaItemCount
+                // player.player.replaceMediaItems seems to stop playback so we
```
nift4: replaceMediaItems probably does but replaceMediaItem without s does not, update current song with repalceMediaItem too

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
```diff
@@ -0,0 +1,582 @@
...
+        )
+
+        val plr = player.endedWorkaroundPlayer!!
+
+        if (mq == null || mq.queue.isEmpty()) {
+            plr.setMediaItems(ArrayList())
+            return null
+        }
+
+        // I have no idea why this value gets reset to 0 by the end... but ig this works
```
nift4: ?

nift4 (changes requested):

app/src/main/res/values/strings.xml
```diff
@@ -503,4 +503,5 @@
     <string name="actions_query_shuffle_specific">shuffle $item_name</string>
     <string name="please_allow_to_delete">Press allow to delete files</string>
     <string name="choose_sd">Choose %s</string>
+    <string name="settings_mq_preview">Enable multiple queues preview</string>
```
nift4: try to avoid mentioning preview, instead use strings that can be reused once feature is stable, to avoid double work for translators

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
mikooomich: For future uses. Should i remove it for the time being?

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
mikooomich: The currently playing media item can be at any index, and it needs to become the first index.. I'm not sure how replaceMediaItem would work better than what I current have.

nift4: I mean in addition to what you have, as replaceMediaItem updates metadata without interruptting playback

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
nift4: having smth for future use is ok if you know for what use, but atleast for me its not really clear what its useful for

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
```diff
@@ -67,4 +74,25 @@ class EndedWorkaroundPlayer(exoPlayer: ExoPlayer) : ForwardingSimpleBasePlayer(e
...
+        startPositionMs: Long
+    ): ListenableFuture<*> {
+        val mq = queueBoard.addQueue(
+            title = "wip ${System.currentTimeMillis()}",
+            mediaList = mediaItems,
+            mediaItemIndex = startIndex,
+            startPositionMs = startPositionMs
+        )
+        queueBoard.commitQueue(mq)
+        return Futures.immediateVoidFuture()
```
nift4: why does this not call super?

mikooomich:
Hello. It has been... a while. I don't remember if I made all the requested changes, so if i missed anything plz let me know.

Known issues I will take care of... soon
- queue title
- wonky queue list close animation
- play buttons from artists/albums etc page do not queue properly


Please help me figure out why the song recycler can scroll down but not up. Please please please please

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
mikooomich: commitQueue calls setCurrQueue, which then calls realSetMediaItems, so calling handleSetMediaItems setmediatems just leads to a duplicate call

nift4: I thought we already etablished that queueboard should only own inactive queues and the title of the active queue. that is, in setMediaItems it should copy the _old_ queue into quereboard, not the new one. because if you want to do it this way then you also need to override at least handleAddMediaItems, handleMoveMediaItems, handleReplaceMediaItems, handleRemoveMediaItems, handleSeek which is stupid amount of effort, while if qb doesn't own current queue 1. nothing related to resumption etc needs to change 2. you avoid all of that spagetthi.

app/src/main/java/org/akanework/gramophone/logic/GramophoneExtensions.kt
```diff
@@ -337,6 +349,149 @@ fun MediaController.getAudioFormat(): AudioFormatDetector.AudioFormats =
...
+
+/*
+// TODO: shuffle and repeat mode
+fun MediaController.playQueue(
+    title: String?,
+    mediaList: List<MediaItem>,
+    mediaItemIndex: Int,
+    isOriginal: Boolean
+) {
+    sendCustomCommand(
```
nift4: commented out much?

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -538,11 +553,15 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
                 if (items != null) {
                     if (endedWorkaroundPlayer?.nextShuffleOrder != null)
                         throw IllegalStateException("shuffleFactory was found orphaned")
-                    endedWorkaroundPlayer?.nextShuffleOrder = factory.toFactory()
                     try {
-                        mediaSession?.player?.setMediaItems(
-                            items.mediaItems, items.startIndex, items.startPositionMs
+                        val mq = endedWorkaroundPlayer?.queueBoard?.addQueue(
```
nift4: see above, I think this part shouldnt need to change at all and qb should only manage inactive queue

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -864,6 +891,91 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
...
 
+                SERVICE_QB_GET_INACTIVE -> {
+                    SessionResult(SessionResult.RESULT_SUCCESS).also { res ->
+                        val queueList: List<MultiQueueObject> = qb.getInactiveQueues()
+                        val binder = BundleListRetriever(queueList.map { it.toBundle() })
+                        res.extras.putBinder("allQueues", binder)
+                    }
+                }
+
+                SERVICE_QB_GET_QUEUE -> {
```
nift4: what is get for? can't we just have user call LOAD and then read it? or is there some UI where content of inactive is needed?

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -864,6 +891,91 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
...
+                                    .toFactory()
+                            val shuffleOrder = factory(0, mq.getSize(), endedWorkaroundPlayer!!)
+                            val shuffleIndexesList: List<Int> = shuffledIndices(shuffleOrder)
+                            res.extras.putIntArray("shuffleIndexes", shuffleIndexesList.toIntArray())
+                        }
+                    }
+                }
+
+                /*
+                SERVICE_QB_ENQUEUE -> {
```
nift4: commented out much?

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -897,7 +1009,7 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
                     if (endedWorkaroundPlayer?.nextShuffleOrder != null)
                         throw IllegalStateException("shuffleFactory was found orphaned")
                     if (isForPlayback && items.mediaItems.isNotEmpty()) {
-                        endedWorkaroundPlayer?.nextShuffleOrder = factory.toFactory()
+//                        endedWorkaroundPlayer?.nextShuffleOrder = factory.toFactory()
```
nift4: yeah...no

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
```diff
@@ -0,0 +1,581 @@
...
+                )
+                plr.replaceMediaItems(
+                    startIndex + 1, Int.MAX_VALUE,
+                    mediaItems.subList(startIndex + 1, mediaItems.size)
+                )
+            }
+
+            plr.exoPlayer.setShuffleOrder(seed.toFactory()(mq.startIndex, mq.getSize(), plr))
+        } else {
+            Log.d(TAG, "Seamless is not supported. Loading songs in directly")
```
nift4, line 436: Remove logs btw

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
```diff
@@ -0,0 +1,581 @@
...
+
+            if (plr.shuffleModeEnabled && mq.shuffleOrder == null)
+                Log.w(TAG, "Shuffle mode is enabled but no shuffle order is provided")
```
nift4: what about not storing that boolean and just making null mean shuffle is off? avoids inconsistency altogether. (shuffle order doesn't need to be stored if shuffle is off because enabling will reshuffle anyway)

nift4 (changes requested):
Sorry don't have time for nested scroll debugging for at least 2 weeks, all I can give you rn is this quick review pass

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
mikooomich: GET returns the queue with shuffle indexes for ui purposes, while LOAD is intended to facilitate the loading of the queue into the player. I don't believe these should be the same command.

app/src/main/java/org/akanework/gramophone/logic/GramophoneExtensions.kt
mikooomich: qb doesn't own the current queue. Yes, a full queue was added into qb via addqueue, but all that data (with the exception of the title) remains untouched and isnt actually used anywhere. All the old data is overwritten anyways with the latest from the player when  setmediaitems is called again, so there is no spaghetti required.

I see how that commitQueue nonsense is redundant, so I'll change that. I'll null out the info in addQueue to prevent that info from creeping in in the future, and also call super.

Now (https://github.com/FoedusProgramme/Gramophone/pull/865/commits/fe511a1ca3f5655140dcaf55c89ba4c895d77a5b) it works like this for when setmediaitems is called: handleSetMediaItems -> addqueue adds skeleton queue to qb, -> commitqueue facilitates the active/inactive swap within qb (without its own setmediaitems) -> super.handleSetMediaItems 

And then for user initiated queue swaps: commitqueue facilitates the active/inactive swap within qb (with realSetMediaItems) -> super.handleSetMediaItems. You cant call setmediaitems again

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
mikooomich: I'd prefer not. I mean, I'm willing to guard it behind a flag, but I wish to keep some debug logs around

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
mikooomich: Alright

app/src/main/java/org/akanework/gramophone/logic/QueueBoard.kt
nift4: flag is fine i was just concerned about runtime logspam

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
nift4: UI purposes as in?

The UI is just the dropdown thingie where the user can select a queue. The user hence has no way to view an inactive queue without activating it in UI, no? Hence, I can't think of any UI purpose that would require a LOAD

app/src/main/java/org/akanework/gramophone/logic/GramophoneExtensions.kt
nift4: > remains untouched and isnt actually used anywhere

should be pretty obvious why thats a bad idea, you can just save the current queue's title in extra member field or something (note: i did not check latest commit)

> And then for user initiated queue swaps: commitqueue facilitates the active/inactive swap within qb (with realSetMediaItems) -> super.handleSetMediaItems. You cant call setmediaitems again

Yes, that sounds reasonable

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
mikooomich: No. When you click on the queue, it previews it, hence GET is used. LOAD is when you click the load queue Button to activate it.

app/src/main/java/org/akanework/gramophone/logic/GramophoneExtensions.kt
```diff
@@ -503,6 +635,16 @@ fun WindowInsetsCompat.clone(): WindowInsetsCompat =
...
     })
 
+fun Context.supportsWideScreen() : Boolean {
+    val config = resources.configuration
+    return config.screenWidthDp >= 780
+}
+
+fun Context.isTablet() : Boolean {
+    val config = resources.configuration
+    return config.smallestScreenWidthDp >= 780
```
nift4, line 645: sw600dp is tablet i think

nift4:
btw is it on purpose that you undrafted it without removing [WIP]

app/src/main/java/org/akanework/gramophone/logic/GramophoneExtensions.kt
mikooomich: Yes ik, but the UI is cramped at 600. Playing around with the value rn. I'm the future i'd like this do either 1) UI based on screen size only or 2) based on device type. Haven't decided yet

mikooomich:
@nift4 Should be ready now. If there are any changes you'd like, please let me know. Otherwise, give me the go ahead and i'll handle merging (via merge commit)

mikooomich:
Hi. I fixed the stupid nested scrolling issue. It should be ready. There should be no behaviour changes when the compile time flag is set to false.


Other stuff and polishing shall come later. This pr has hurt me in way too many ways. Meow.

app/src/main/java/org/akanework/gramophone/logic/ui/QueueRecyclerView.kt
```diff
@@ -0,0 +1,32 @@
...
+    constructor(context: Context, attributeSet: AttributeSet?, defStyleAttr: Int)
+            : super(context, attributeSet, defStyleAttr)
+
+    constructor(context: Context, attributeSet: AttributeSet?) : super(context, attributeSet, 0)
+
+    constructor(context: Context) : super(context, null)
+
+    private var lastYPos = 0f
+
+    override fun onInterceptTouchEvent(e: MotionEvent): Boolean {
```
nift4, line 17: I don't think touch event interception is how this is intended to be done? Nested scroll has pre/post scroll specifically for priority

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
```diff
@@ -14,11 +21,27 @@ import org.akanework.gramophone.logic.utils.CircularShuffleOrder
...
+class EndedWorkaroundPlayer(
+    exoPlayer: ExoPlayer,
+    val queueBoard: QueueBoard
+) : ForwardingSimpleBasePlayer(exoPlayer),
     Player.Listener {
 
     companion object {
         private const val TAG = "EndedWorkaroundPlayer"
+
+        fun queueWithTitle(mediaItems: List<MediaItem>, mqTitle: String?): List<MediaItem> {
```
nift4, line 33: This api should go elsewhere

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
```diff
@@ -67,4 +90,29 @@ class EndedWorkaroundPlayer(exoPlayer: ExoPlayer) : ForwardingSimpleBasePlayer(e
...
+    ): ListenableFuture<*> {
+        val defaultQueueTitle = queueBoard.context.getString(R.string.unknown_playlist)
+        val firstMediaItem = mediaItems.firstOrNull()
+        val mqTitle = firstMediaItem?.mediaMetadata?.extras?.getString("mq_title")
+
+        val mq = queueBoard.addQueue(
+            title = mqTitle ?: defaultQueueTitle,
+            mediaList = ArrayList(),
+            mediaItemIndex = C.INDEX_UNSET,
+            startPositionMs = C.TIME_UNSET,
```
nift4, line 113: are you making a dummy queue every time? is that really the best way?

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
```diff
@@ -67,4 +90,29 @@ class EndedWorkaroundPlayer(exoPlayer: ExoPlayer) : ForwardingSimpleBasePlayer(e
...
+        startIndex: Int,
+        startPositionMs: Long
+    ) = super.handleSetMediaItems(mediaItems, startIndex, startPositionMs)
+
+    override fun handleSetMediaItems(
+        mediaItems: List<MediaItem>,
+        startIndex: Int,
+        startPositionMs: Long
+    ): ListenableFuture<*> {
+        val defaultQueueTitle = queueBoard.context.getString(R.string.unknown_playlist)
```
nift4, line 105: String resources should optimally not be resolved before use, because they won't update when user changes UI language. Title could be either resource ID or string (if derived from user's library), for example

app/src/main/java/org/akanework/gramophone/ui/SongPickerActivity.kt
```diff
@@ -13,6 +14,7 @@ class SongPickerActivity : PickerActivity<MediaItem>() {
     override fun makeAdapter() =
         SongAdapter(
             null,
+            flowOf("MISSING TITLE: (SongPickerActivity)"),
```
nift4, line 17: add a TODO

app/src/main/java/org/akanework/gramophone/ui/fragments/PlaylistEditFragment.kt
```diff
@@ -80,6 +83,7 @@ class PlaylistEditFragment : BaseFragment(false) {
                 onRequest(it.resultCode)
             }
         theItem = MutableSharedFlow(replay = 1)
+        qTitle = theItem.map { it?.title ?: "MISSING TITLE (PlaylistEditFragment)" }
```
nift4, line 86: add a todo for things like these

nift4 (commented):
I'm merging it so that we can iterate better (and without a billion merge conflicts) but please fix these

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
mikooomich: No. 

Just passes the title, etc. the addqueue will edit exusting queue, or create a new one if one with same title doesn't exist. Then addqueue returns that queue 

What would you have done?

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
mikooomich: Why would that matter here? Isn't Title is retrieved on every function call

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
nift4: add a seperate field in queueboard for active queue's title which is being read

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
nift4: imagine the user creates a queue, then changes language. the already existing queue will have name in old language because it was created before language change

app/src/main/java/org/akanework/gramophone/ui/SongPickerActivity.kt
mikooomich: Ah, right, these. I forgot to ask what I should put as title for these, not very accessable in ui

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
mikooomich: I don't think that's the way to go. Wouldn't that be storing title in 2 different places? And also not change much. I'm a bit confused why you would want that

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
mikooomich: ofc it can be done but... I think we should just snapshot the title as queues are of the library/playlist/etc. as they queue is just a snapshot of that. I wouldn't change it to a nullable field for the sake of changing it... the user has to A) Play an undefined queue, which basically won't exist in the future unless dev messes up AND  B) changes the language AND C) Cause a bug worth fixing at all. I think it's more likely i win the lottery tomorrow...

nift4: unknown playlist isnt the only string you resolve fwiw, you also have stuff like Songs: %s which can change, but it is quite unimportant in the grand scheme of things i guess

PR #886, opened 2026-04-14
Migrate build configuration to Gradle Version Catalogs

iscle:
This PR migrates the built do use Version Catalogs for easier library management.

No actual versions were changed, they have all been kept the same as before.


AkaneTan:
Hi, there's a reason that we're not using version catalog on gramophone, that is it is only convenient in large projects containing multiple libraries which require unified dependency version. Gramophone is not the suitable case apparently

iscle:
Thanks for the reply! I think that even for small projects it's a good QoL improvement. It removes clutter from the build.gradle.kts file and helps keeping everything organized in its own file.

But of course, this comes down to developer preference! Feel free to close the pr if it's not something you're interested in :)

Issue #577, opened 2025-07-15
[BUG] playback resumption after reboot has no album cover


nift4:
https://github.com/androidx/media/issues/2331

nift4:
Well, I'll need a proxy content provider or make it load the Bitmap?

nift4:
Prioritizing as this bug also breaks Bluetooth AVRCP, causes a _lot_ of log spam, and is a regression.

Issue #593, opened 2025-07-22
[FR] Android Auto app template implementation


nift4:
related to #93 
also see https://developer.android.com/training/cars/apps/media

Issue #881, opened 2026-03-30
[FR] Databases design plan


mikooomich:
A divine plan for playback tracking and tag cache where we will pray that there will be no edge cases.

## Db design
(Using room + custom queries as needed)

Tags cache db
- **columns**: chromaprint(?), generation (before R: mtime), artist, title, path (from volume root), volume name, etc...
  - if any of mtime/generation or size changes, the entry is rescanned
  - aggressively prune old tags (for files that no longer exist), except if they're on a volume which android thinks will come back soon
- primary source of tags
  - Mediastore data will be used as fallback if the song doesn't exist in the tag cache 
  - if the file's tags are scanned while the app is open, the tags should update in every list automatically
  - TODO: maybe also update the current playlist?

Play counts/history db
- Song info table
  - **columns**: chromaprint (not entry id), fallback attributes: file name and title + artist + album + year
  - Null chromaprint until we have it, then Merge with existing rows if necessary 
- History table
  - **columns**: id (ref), datetime, listen duration

## Misc
- How to chromaprint impl
  - [Modernish](https://github.com/cy745/fpcalc) version of [the original](https://github.com/QuickLyric/fpcalc-android).
  - [Hi](https://httpraccoons.com/428)
- Read chromaprint tags from metadata directly

nift4:
https://developer.android.com/reference/android/provider/MediaStore#getRecentExternalVolumeNames(android.content.Context)

"These volume names are not currently mounted, but they're likely to reappear in the future, so apps are encouraged to preserve any indexed metadata related to these volumes to optimize user experiences."

reversing the conclusion, if a volume is not in that list, we should discard its tag cache data.

edit: also got the request to be able to exclude albums or songs from history entirely

nift4:
https://github.com/nbeaver/chromaprint-playlists

nift4:
Got request to support Composer tags for classical music and a Composer tab in UI, hmm

PR #842, opened 2026-02-17
feat: Android Auto Media Browsing Support

ghhccghk:
Adds media browsing functionality to Android Auto.

Key changes include:
- Implemented `onGetLibraryRoot` and `onGetChildren` in `GramophonePlaybackService` to provide the media library browsing structure, currently supporting only the “current playback queue”.
- Added `GramophoneAlbumArtProvider` to supply album art to Android Auto via `ContentProvider`, resolving cover art display issues on Android Auto.
- Refactored `dumpPlaylist` logic and migrated it to the new `PlayerListHelp` utility class for reuse between `Service` and `UI`.
- Registered `GramophoneAlbumArtProvider` in `AndroidManifest.xml` and officially enabled Android Auto functionality.

Translated with DeepL.com (free version)

PR #890, opened 2026-04-28
Add Android Auto Support

### LINKED ISSUE #93: [FR] Android Auto
Enable Android Auto integration for Gramophone to enhance the in-car music experience. This feature would allow users to seamlessly access and control Gramophone's music playback directly from their Android Auto-enabled car infotainment systems. Integration with Android Auto will not only provide a safer and more convenient way for users to enjoy their favorite music while driving but also extend Gramophone's reach to a wider audience of Android Auto users.

banksio:
Hi! I use Gramophone and love it, so I wanted to try my hand at a PR to add support for Android Auto. I've tested this with the Desktop Head Unit and my own car, and it works great.

---

**I've made the following changes to implement robust Android Auto support to Gramophone:**

- Implemented full `MediaLibraryService` support to get Android Auto working. I surface the **Albums**, **Artists**, **Songs**, and **Playlists** tabs (AA limits to 4 tabs). The queue behaviour is consistent between tapping an item in the mobile UI and the Android Auto UI.
- Refactored the media library browsing logic into a separate `LibraryTreeLoader` class, neatly separating the tree-building logic from the main playback service.
- Polished the UI experience by ensuring empty categories and playlists (like "Recently Added" or "Favourites" when empty) are properly marked as non-tappable in the Android Auto interface.
- Added `setPeriodicPositionUpdateEnabled(false)`. This is a workaround for a known Media3 bug (androidx/media/issues/2192) and doesn't negatively impact the player or UI in any meaningful way from my testing.  
- Fully integrated with the new Coil 3 artwork pipeline that was recently merged upstream, ensuring album art loads smoothly and natively in Android Auto without requiring a custom `ContentProvider`.

---

*(Note: The previous custom artwork caching changes and the Wear OS album art workaround have been removed from this PR. The artwork infrastructure is now handled seamlessly by the upstream Coil 3 changes, and I will open a separate PR to discuss the Wear OS fix independently to keep this PR focused purely on Android Auto.)*

Also note that I did not go down the template route (#593), as [they are in beta and currently not able to be published publicly to the Play Store](https://developer.android.com/training/cars/apps/media#:~:text=Templated%20media%20apps%20are%20in%20beta). Therefore it is much easier to use the current `MediaLibrarySession` for now. Furthermore, Gramophone doesn't require the use of any custom flows for Android Auto as it is a fairly simple media player.

Closes #93.

---

Thanks for the app, and keen to hear your thoughts on my PR and if any changes are recommended! I can provide screenshots if requested also.

Please feel free to test in your own vehicle or with the [Desktop Head Unit](https://developer.android.com/training/cars/testing/dhu).


mikooomich:
#842 also implements AA, no idea why that pr isn't showing up in the pull request screen though

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -501,6 +547,7 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
                     )
                 )
                 .setSystemUiPlaybackResumptionOptIn(Build.VERSION.SDK_INT >= Build.VERSION_CODES.S)
+                .setPeriodicPositionUpdateEnabled(false)
```
nift4, line 570: The media3 issue you linked about this said that they assume it's fixed in newer Android Auto app versions. Which version did you try? (Also, in such cases it's helpful to leave a one-line comment stating why this was done)

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -394,10 +408,42 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
...
+                val original = super.getCurrentTimeline()
+                return object : androidx.media3.exoplayer.source.ForwardingTimeline(original) {
+                    override fun getWindow(windowIndex: Int, window: androidx.media3.common.Timeline.Window, defaultPositionProjectionUs: Long): androidx.media3.common.Timeline.Window {
+                        super.getWindow(windowIndex, window, defaultPositionProjectionUs)
+                        window.mediaItem = convertItem(window.mediaItem)
+                        return window
+                    }
+                }
+            }
+        }
```
nift4: I'm not sure if wrapping these media items at this level is the best solution. The gramophoneAlbumCover:// custom scheme thing doesn't only cause issues with Android Auto, there is also #577 (playback resumption), Bluetooth AVRCP and https://developer.android.com/media/optimize/mct with which you can see the issue. So my plan was to migrate the entire app's data structures away (see Reader.kt) from that custom scheme and instead add a Coil hook that bypasses the provider if we load the uri in-process

app/src/main/java/org/akanework/gramophone/logic/utils/ArtCacheManager.kt
```diff
@@ -0,0 +1,152 @@
...
+        val mimeFile = File(cacheDir, "art_$key.mime")
+
+        if (cacheFile.exists()) {
+            cacheFile.setLastModified(System.currentTimeMillis())
+            val mimeType = if (mimeFile.exists()) mimeFile.readText() else "image/jpeg"
+            return ArtResult(cacheFile, mimeType)
+        }
+
+        // Materialize
+        val artStream = ArtResolver.openResourceStream(context, resource, size) ?: return null
```
nift4: I disabled coil's disk cache because (if enhanced loading is disabled):
1. MediaStore has its own disk cache, and it provides album covers with 50% of screen resolution as quality (so it's fine for every thumbnail use)
2. The only time we need a higher-quality cover is in the "now playing" screen, and that is only needed for one song at a time, so we have basically no advantage of caching that. So we get that from ThumbnailUtils directly (which checks both embedded covers and .jpg files in the folder)
3. If ThumbnailUtils can't find embedded cover and we don't have picture permission (-> can't see .jpg file), we have to ask MediaStore even in high quality case because there may be a .jpg file we can't see. It will just fail if there's no cover

Enhanced album cover setting then has two distinct functionality changes:
1. Due to being granted picture permission, ThumbnailUtils can now find .jpg files automatically. Thus, we no longer need to ask MediaStore as fallback because we now have same set. This doesn't need to be cached because it's still only the high quality case
2. We can also ourselves look if the album has a different cover than any of the songs (I don't remember why I added it to be honest, but I still remember someone gave me an album where every song had a different cover and it probably was a result of that). This one should probably be cached

The high quality version of the cover is provided to Android Auto using MediaSession metadata as Bitmap by media3, it doesn't have to be queried using the content provider. So the content provider can be a proxy to MediaStore except for the case where album cover is different from song cover. In this last case only it would need custom loading logic and caching.

As you can see there is only one edge case we need to cache, and I'm still wondering whether we could reuse the Coil implementation for that instead of doing it in a custom way.

(Also, I know, none of this is documented... I really should, but well...)

app/src/main/java/org/akanework/gramophone/logic/GramophoneLibrarySessionCallback.kt
```diff
@@ -0,0 +1,395 @@
...
+ * Handles the media library browsing logic for [GramophonePlaybackService].
+ * Extracted from the service to improve maintainability and separate UI/tree-building logic.
+ */
+class GramophoneLibrarySessionCallback(
+    private val context: Context,
+    private val app: GramophoneApplication,
+    private val lifecycleScope: LifecycleCoroutineScope,
+    private val convertItem: (MediaItem) -> MediaItem,
+    private val delegate: SessionDelegate
+) : MediaLibrarySession.Callback {
```
nift4: I think your delegate pattern is a really good idea, but I'd switch that order (the service remains the implementation of the actual Callback class, and this one is called by the service for everything related to library loading)

nift4 (commented):
Hi, thanks for the PR! It's currently a busy time in my life, so I will need 2-3 weeks until I can review it in detail. Thanks for your understanding!

Until then, I'll do a surface-level review of these changes with some questions. Note that I haven't tried or looked into detail a lot yet, it's only from a quick look.

> I surface the Albums, Artists, Songs and Playlists tabs (AA limits to 4 tabs). 

You could have a "More" tab as fourth tab that offers the other tabs. Also, I think it would be nice if it would respect the tab order setting from Appearance settings, with which user can customize the most important tabs in the phone UI.

nift4:
Oh, also:

> Reduce Duplication of Work: If Enhanced album cover loading is on, the album art file found in the filesystem will be used for all songs in an album by default, rather than loading and caching the art from each song.

As I said above, this was on purpose, as I got sent an album where each song had a different cover in a bug report some time ago.

> This provider is limited to Android Auto only for now via the com.google.android.finsky.permission.GEARHEAD_SERVICE permission, for privacy.

the content://media/audio/albumart/X uris on platform side actually don't enforce any permission, and that's because ecosystem support for that is just not there...

https://cs.android.com/android/platform/superproject/+/android-latest-release:packages/providers/MediaProvider/src/com/android/providers/media/MediaProvider.java;l=10363;drc=7fa7a8fad38b5cda571add687d153b4a998d6348

```
        // check access only for image and video thumbnails
        // audio thumbnails have many legacy paths that we could break by checking for access
        // and it doesn't reveal much of data that could be a risk
```

I don't think we'll get away with any kind of permission checks while making apps like Musixmatch, Android Auto, Bluetooth AVRCP, last.fm scrobblers, etc work. And even if we do check permission, an attacker could always just use the platform version of the Uris and access them without any check, so it would be a bit pointless.

> Also note that I did not go down the template route

That's fine, the plan was indeed to first to a browser service as that is also used for Bluetooth AVRCP

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
banksio: I'm on `16.6.661444-release`.

This one took me a good couple of days of debugging/testing. I am assuming that it either was fixed and regressed again, or was never fixed. I tested a few apps. YouTube Music, Spotify and Tidal don't seem to have the bug (though, they might be using the workaround...), but https://github.com/Ghost-Applications/gizz-tapes does have it (and does not use the workaround).

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
banksio: I understand and I see where you are going with migrating off the custom schemes in `Reader.kt`. I think it would simplify the code a lot.

I'll play around with refactoring the code to achieve that migration and let you know how I get on.

app/src/main/java/org/akanework/gramophone/logic/utils/ArtCacheManager.kt
banksio: Thanks for explaining. I didn't know the quality that `MediaStore` provided or that `ThumbnailUtils` will pull the folder `.jpg`.

Yep, it doesn't actually make sense to duplicate MediaStore's cache.

To explain why I did it this way; I was considering the case of the user storing a ridiculously large album art in a directory (or embedded in the audio file). Perhaps we should resize everything to a certain size (if it's above, say, 100% of the devices' screen resolution) even for the high quality art. And then cache that downsized art.

Gramophone could handle that resizing, or it could treat it as the user's problem with the mindset that they should not have such big album art. Both are valid, I'd say.

Anecdotally I have a 7MB cover in one of my albums. I noted that Wear OS fails to load the album art if Enhanced loading is on, likely due to that art being so large. Ideally this PR could set up the art infrastructure such that a future PR could fix that Wear OS support as well.

banksio: And yes I'll revert the removal of the per-song album art as default since you mention it is on purpose. We could have a setting for it maybe, as it does reduce a lot of work on the CPU if the user doesn't require it?

banksio:
Hey @nift4, thanks for the detailed comments and explanations! This really helps me to understand your goals for the code. No worries on the delay, life is busy sometimes. I'd love to help out where I can. I'll address your comments and work on the code in the PR over the next couple of weeks.

1. I agree with the tab ordering matching the user's setting, and having a  _More_ tab. I thought about doing that but left it out for some reason. I'll implement it this week.
2. I'll just remove the permission line from the content provider since it doesn't seem to be respected anyway.
3. (Other comments replied in-threads.)

Thanks again for the detail, I've learnt a lot from reading your replies!

app/src/main/java/org/akanework/gramophone/logic/utils/ArtCacheManager.kt
nift4: > Perhaps we should resize everything to a certain size (if it's above, say, 100% of the devices' screen resolution) even for the high quality art. And then cache that downsized art.

The reason why I did not is actually something you found out 2 paragraphs later: it's not only displayed on the phone screen, it might also get synced to WearOS, Chromebook / Link to Windows, etc, and I don't know how good of a quality they need.

WearOS failing to display the cover because it's _too_ big is quite stupid and wasn't what I would expect....

nift4: > We could have a setting for it maybe, as it does reduce a lot of work on the CPU if the user doesn't require it?

Do you use audio files with no embedded art + sidecar image file (this one can be detected/optimized to some extent)? Or do you use files that have embedded art individually, but they just have all the same file?

banksio: I currently use files that all have the same embedded art, then I also put the sidecar `cover.jpg` (which is slightly higher quality) in the folder for redundancy.

nift4: I don't think this is a great fit for a setting, as both consequences (wrong album art displayed, or high CPU usage) are very much not obvious. I would rather leave it has high CPU usage for now, and when Gramophone gets its own tag cache (see #881) it can be checked if multiple songs use the same cover as part of building that cache. Then the optimization will be invisible to the user and won't require toggling a setting.

banksio: That's fair, I see your point on the tag cache.

I'll look to remove the cache for now then, and lean on the Coil cache for the single case you mention.

nift4: FYI; I pushed a commit that uses a "ThumbnailUtilsCompat" wrapper to replace ThumbnailUtils (with the same API surface), because ThumbnailUtils only gained the API we want in Android Q, in order to make sure that Android <=P and >=Q can use the same code flow. It did cause a conflict but it shouldn't be much work to adapt your WIP branch.

banksio:
Just to let you know, I'm still planning to get around to this, I've just been a bit busy lately. I've sorted the tab ordering in my fork but haven't pushed it to the PR branch yet. Watch this space!

nift4:
No worries, I'll miss my own deadline for the 1.1 release (May 20th, one year 😭 after the last one) anyway given there is still enough work to be done. Depending on how much time that takes on my side (my replacement deadline is first week of June), AA might make the cut for 1.1 feature release, but even if not - taking so long between releases harms the project, so I decided to make more releases with less content (whatever is done in time), so we can just go for the next one.

Well all that is to say, there is no hurry :)

(The only thing that depends on how fast you are is that I would like to fix bug #577 before the 1.1 release - as you can see by the "high priority" label - so I might end up adding a half baked albumart ContentProvider myself if the PR isn't done in time, which might result in more conflicts - sorry about that in advance)

banksio:
Just as an update, this is almost ready now. The only outstanding comment is to look at the custom schemes `gramophoneAlbumCover://` etc.

- Android Auto now respects the user's tab order preference and shows a More tab when necessary. _Filesystem_ is excluded from this, as that's a fairly complex tab, and probably not really one we should display in a car.
- I've reversed the delegate pattern as you mentioned above.
- Coil is the sole image caching mechanism. It will cache everything in memory as required, temporarily, to ensure app responsiveness, and cache album art on disk where it is different to the song art.
- Song artwork is always preferred over album art.
- Wear OS album art is fixed. The app will resize album art to 512x512 for media session metadata, alongside the provider URI, so both a lower-res and high-res option are available. Wear OS or any other apps that can't follow the URI will grab the 512x512 and everything else will get the high-res image from the content provider URI.

I'm running this build as my daily, so am using that for testing and will come back to resolve any issues I find.

nift4:
fyi, you dont have to fix everything codescene complains about. i didnt configure the thresholds down yet so its more complainy than needed.

nift4:
Sorry, I dropped the CodeScene cleanup commits because they make it confusing to review.

nift4:
I merged a large part (but not all of) the album art changes in 0735b85da17956df5f48255f51ec1c6e6ba4d0bb / 10586523a427cc3a4826e3edf1dba2c7f0f280a3 to fix #577. (Specifically, I skipped the Wear OS workaround for now as I think it'd be best if that gets more detailed review, as byte[] of images is easy to make memory mistakes with, and I wonder if it may be fruitful to report the bug to Google instead so that the Wear OS app downscales images instead of failing.)

I did end up changing some things because while testing your nicely prepared pull request, I noticed a few of my former assumptions were wrong:
* I thought coil3 supports disk-caching images scaled to a smaller size - it does not! The disk cache is only used for http(s) images in coil by default, for any other usecase, extra code is needed. See https://github.com/coil-kt/coil/issues/2926#issuecomment-2843908440
* We also don't need to re-compress() Bitmaps from albumart Uris or ThumbnailUtils all the time in AlbumArtProvider, because there's some fairly undocumented APIs to get the thumbnails as JPEG data (which we can pass through unmodified)
* The assumption we need a disk cache at all also turned out to be wrong, MediaStore also provides disk cache capability for images using ContentResolver.loadThumbnail().

Can you:
1. work on un-tangling the remaining Android Auto changes (I think it shouldn't be hard, given that the changes are seperated nicely in seperate files) and force-push, so that the PR is mergable again on latest commit?
2. send the Wear OS related changes again in a seperate PR so it can be discussed or debugged seperately?

If you don't understand the mess I made ;) I can also look at untangling of Android Auto myself, but it might take a while...

banksio:
Hi @nift4 I've rebased the Android Auto changes onto the latest upstream/beta to resolve the merge conflicts and hopefully the branch can now merge cleanly.

* Cleanly resolved the GramophonePlaybackService.kt conflicts, retaining the new albumIdFallback and mapMediaItemsForFavorites logic.
* Dropped the legacy artwork caching commits from this branch, as the new GramophoneAlbumArtProvider and Coil 3 setup handles this now.
* Fixed the onAddMediaItems signature (MutableList -> List) to match the latest Media3 version.
* Refined LibraryTreeLoader: Empty playlists (like 'Recently Added' or 'Favorites' when empty) are now properly marked as non-tappable in the Android Auto interface (isPlayable = false, isBrowsable = false).

I've removed the Wear OS album art workaround as requested. I'll open a separate PR for that later.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
banksio: Just tested on `17.2.662634-release` and this issue is still present, so this workaround is still required.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -191,8 +194,30 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
...
     val endedWorkaroundPlayer
-        get() = mediaSession?.player as EndedWorkaroundPlayer?
+        get() = internalPlayer
+
+    private lateinit var libraryTreeLoader: LibraryTreeLoader
+
+    private fun convertMetadata(metadata: MediaMetadata, albumIdFallback: Long? = null): MediaMetadata {
+        val artworkUri = metadata.artworkUri ?: return metadata
+        val scheme = artworkUri.scheme
+        if (scheme == "gramophoneSongCover" || scheme == "gramophoneAlbumCover") {
```
nift4: I think this is not needed anymore, and the entire convertItem can be removed

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -414,7 +439,7 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
                                 .apply {
                                     val config =
                                         prefs.getStringStrict("offload", "0")?.toIntOrNull()
-                                    if (config != null && config > 0 && Flags.OFFLOAD) {
+                                    if (config != null && config > 0) {
```
nift4: Please don't remove the Flags.OFFLOAD

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -433,10 +458,41 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
         player.exoPlayer.setShuffleOrder(CircularShuffleOrder(player, 0, 0, Random.nextLong()))
         lastPlayedManager = LastPlayedManager(this, player)
         lastPlayedManager.allowSavingState = false
+        internalPlayer = player
+
+        val sessionPlayer = object : androidx.media3.common.ForwardingPlayer(player) {
```
nift4: I think this ForwardingPlayer isn't needed anymore as convertItem/convertMetadata no longer does anything

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -1447,6 +1485,14 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
 
     override fun onEvents(player: Player, events: Player.Events) {
         super<Player.Listener>.onEvents(player, events)
+        if (events.containsAny(
+                Player.EVENT_REPEAT_MODE_CHANGED,
+                Player.EVENT_SHUFFLE_MODE_ENABLED_CHANGED,
+                Player.EVENT_TIMELINE_CHANGED
+            )
+        ) {
+            refreshMediaButtonCustomLayout()
```
nift4: are the existing callers not enough?

app/src/main/AndroidManifest.xml
```diff
@@ -268,6 +266,12 @@
                 android:resource="@xml/file_paths" />
         </provider>
 
+        <provider
+            android:name=".logic.GramophoneAlbumArtProvider"
```
nift4: I think this is now a duplicate

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -1740,4 +1740,87 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
...
+                    try {
+                        val expanded = future.get()
+                        val mapped = mapMediaItemsForFavorites(expanded.mediaItems).map { convertItem(it)!! }
+                        completer.set(mapped)
+                        val startIdx = expanded.startIndex
+                        if (startIdx != null && mapped.isNotEmpty()) {
+                            handler.post {
+                                mediaSession.player.setMediaItems(
+                                    mapped,
+                                    startIdx,
```
nift4: If you need to provide a starting index, onAddMediaItems is the wrong callback. 

There is the onSetMediaItems callback which is very similar (and in the default implementation, which is what's used right now, is calling onAddMediaItems as fallback), but supports starting index directly.

You nonetheless need an implementation of onAddMediaItems, but that then shouldn't assume it can change the starting index. Because if you override onSetMediaItems to NOT call onAddMediaItems (but instead do custom logic with the starting index), then onAddMediaItems will only be used in cases where a media item should actually be added (for example, "add to end of queue" or "play next"). Fwiw, that can't be triggered from Android Auto I believe, Android Auto always only calls onSetMediaItems, but there is for example https://developer.android.com/media/optimize/mct - and certain Google Assitant voice commands - that can trigger onAddMediaItems which is why we need it anyway.

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,382 @@
...
+            is Date -> createFolderItem(
+                "date_${item.id}", item.title ?: context.getString(R.string.unknown_year), MediaMetadata.MEDIA_TYPE_FOLDER_MIXED,
+                subtitle = context.resources.getQuantityString(R.plurals.songs, item.songList.size, item.songList.size),
+                isPlayable = item.songList.isNotEmpty(), isBrowsable = false, artworkUri = null
+            )
+            is FileNode -> createFolderItem(
+                "folder_${item.folderName}", item.folderName, MediaMetadata.MEDIA_TYPE_FOLDER_MIXED,
+                subtitle = context.resources.getQuantityString(R.plurals.items, item.folderList.size + item.songList.size, item.folderList.size + item.songList.size),
+                isPlayable = item.songList.isNotEmpty() || item.folderList.isNotEmpty(), isBrowsable = false
+            )
```
nift4, line 123: could all of these domains be browsable _and_ playable?

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,382 @@
...
+                        else -> it.id?.toString() == idStr
+                    }
+                }
+                Triple(playlist?.songList ?: emptyList(), 8, true)
+            }
+            else -> Triple(emptyList(), -1, false)
+        }
+        if (adapterType == -1) return emptyList()
+        val sorter = Sorter(SongAdapter.MediaItemHelper, null, if (naturalOrder) Sorter.Type.NaturalOrder else null)
+        return sortList(songs, adapterType, sorter)
```
nift4, line 165: I think the adapter type wasn't the best design from my side, but so far it was ok because the condition and the magic numbers were directly next to each other. Now that the magic numbers are needed to be used elsewhere... maybe add some constants for the numbers that are relevant to AA as well, and use the constants both here and in FileOpUtils, to make it clearer why some adapter type is some number

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    private suspend fun searchForMediaItem(item: MediaItem): List<MediaItem> {
+        val text = item.requestMetadata.searchQuery?.trim() ?: ""
+        val list = app.reader.songListFlow.first()
+        val sortedList = sortList(list, 7, Sorter(SongAdapter.MediaItemHelper, null))
+        return if (text == "") sortedList else sortedList.filter {
+            it.mediaMetadata.title?.contains(text, true) == true ||
+            it.mediaMetadata.albumTitle?.contains(text, true) == true ||
+            it.mediaMetadata.artist?.contains(text, true) == true
```
nift4: I know the TODOs in the original version were ugly but I think it'd make sense to keep them around because they aren't solved

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,382 @@
...
+    private suspend fun searchForMediaItemSync(query: String): List<MediaItem> {
+        val text = query.trim()
+        val list = app.reader.songListFlow.first()
+        val sortedList = sortList(list, 7, Sorter(SongAdapter.MediaItemHelper, null))
+        return if (text == "") sortedList else sortedList.filter {
+            it.mediaMetadata.title?.contains(text, true) == true ||
+            it.mediaMetadata.albumTitle?.contains(text, true) == true ||
+            it.mediaMetadata.artist?.contains(text, true) == true
+        }
+    }
```
nift4, line 325: is this an 1:1 duplicate of the other one?

nift4 (changes requested):
I tried it myself in the DHU and it's already pretty nice, thanks for working on this. The devil is in the details; I left a few new comments, but they should be relatively straightforward to fix.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
banksio: Apologies, that got lost in the conflict resolution! Will restore.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
banksio: In hindsight, they should be. I'll test.

banksio: They are. I'll remove this one.

app/src/main/AndroidManifest.xml
banksio: Removed.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
banksio: Removed.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
banksio: Restored.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
banksio: Removed.

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
banksio: They can, but AA will prioritise the "playable" flag, and simply play the item/folder rather than letting you browse it. I kept it like this to match what YouTube Music does, do tapping an album just plays that album, artist the same etc.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
banksio: Thanks for the detailed information - I understand now. I've updated this code.

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
banksio: Restored.

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
banksio: Yes, refactored.

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
banksio: [Custom browse actions](https://developer.android.com/training/cars/media/create-media-browser/custom-browse-actions) are a thing but are only supported on AAOS right now, so I can't test that.

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
banksio: Added a new library to hold these constants.

banksio:
I've addressed everything individually (hopefully!) and tested. I've also added the new favourite button to the media session.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -813,8 +847,9 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
             if (this.controller?.currentTimeline?.isEmpty == false) {
                 builder.setMediaButtonPreferences(
                     ImmutableList.of(
-                        getRepeatCommand(),
-                        getShufflingCommand()
+                        getFavoriteCommand(),
```
nift4: Let's add the heart button to the end for now, because it otherwise removes repeat button from phone-side notification

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -191,8 +196,12 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
     private val internalPlaybackThread =
         HandlerThread("ExoPlayer:Playback", Process.THREAD_PRIORITY_AUDIO)
     private var mediaSession: MediaLibrarySession? = null
+    private var internalPlayer: EndedWorkaroundPlayer? = null
```
nift4: I think the variable isn't needed anymore

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -345,6 +363,14 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
                     .setDisplayName(getString(R.string.repeat_mode))
                     .setPlayerCommand(Player.COMMAND_SET_REPEAT_MODE, Player.REPEAT_MODE_OFF)
                     .build(),
+                CommandButton.Builder(CommandButton.ICON_HEART_UNFILLED) // not favorite, click will favorite
+                    .setDisplayName(getString(R.string.favorite))
+                    .setSessionCommand(SessionCommand(SERVICE_FAVORITE, Bundle.EMPTY))
```
nift4: Try using the non-custom SET_RATING, then you can avoid adding code in onCustomCommand:
```
.setSessionCommand(SessionCommand(SessionCommand.COMMAND_CODE_SESSION_SET_RATING),
                        HeartRating(true))
```

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -1740,4 +1721,102 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
...
+        return libraryTreeLoader.getSearchResult(query, page, pageSize, params)
+    }
+
+    override fun onAddMediaItems(
+        mediaSession: MediaSession,
+        controller: MediaSession.ControllerInfo,
+        mediaItems: List<MediaItem>
+    ): ListenableFuture<List<MediaItem>> {
+        val future = libraryTreeLoader.addMediaItems(mediaItems)
+        val completion = CallbackToFutureAdapter.getFuture<List<MediaItem>> { completer ->
```
nift4: guava has Futures.transform that can be used to remove roughly half of the lines of code in this function, you can even pass `Dispatchers.Default.asExecutor()` to run it on bg thread

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -1740,4 +1721,102 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
...
+
+    override fun onSetMediaItems(
+        mediaSession: MediaSession,
+        controller: MediaSession.ControllerInfo,
+        mediaItems: List<MediaItem>,
+        startIndex: Int,
+        startPositionMs: Long
+    ): ListenableFuture<MediaSession.MediaItemsWithStartPosition> {
+        val future = libraryTreeLoader.addMediaItems(mediaItems)
+        val completion = CallbackToFutureAdapter.getFuture<MediaSession.MediaItemsWithStartPosition> { completer ->
```
nift4: Again, Futures.transform likely helpful

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,386 @@
...
+            is Playlist -> {
+                val title = when (item) {
+                    is uk.akane.libphonograph.dynamicitem.RecentlyAdded -> context.getString(R.string.recently_added)
+                    is uk.akane.libphonograph.dynamicitem.Favorite -> context.getString(R.string.playlist_favourite)
+                    else -> item.title ?: ""
+                }
+                val icon = when (item) {
+                    is uk.akane.libphonograph.dynamicitem.RecentlyAdded -> "android.resource://${context.packageName}/${R.drawable.ic_default_cover_playlist_recently}".toUri()
+                    is uk.akane.libphonograph.dynamicitem.Favorite -> "android.resource://${context.packageName}/${R.drawable.ic_default_cover_playlist_favorite}".toUri()
+                    else -> null
```
nift4: You can use the first song's cover, that's how the phone UI does it

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
nift4: Oh I didn't realize they were AAOS only, I vaguely remembered something like that but I must've confused it. Then it's fine, thanks.

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,386 @@
...
+            "folders" -> createFolderItem("folders", context.getString(R.string.folders), MediaMetadata.MEDIA_TYPE_FOLDER_MIXED)
+            "playlists" -> createFolderItem("playlists", context.getString(R.string.category_playlists), MediaMetadata.MEDIA_TYPE_FOLDER_PLAYLISTS)
+            else -> null
+        }
+    }
+
+    private fun mapDomainItemToMediaItem(item: Any): MediaItem? {
+        return when (item) {
+            is Album -> createFolderItem(
+                "album_${item.id}", item.title ?: "", MediaMetadata.MEDIA_TYPE_FOLDER_MIXED,
```
nift4: MEDIA_TYPE_ALBUM

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,386 @@
...
+                subtitle = item.albumArtist ?: item.songList.firstOrNull()?.mediaMetadata?.artist?.toString(),
+                artworkUri = item.cover, isPlayable = item.songList.isNotEmpty(), isBrowsable = false
+            )
+            is Artist -> createFolderItem(
+                "artist_${item.title}", item.title ?: "", MediaMetadata.MEDIA_TYPE_FOLDER_MIXED,
```
nift4: MEDIA_TYPE_ARTIST

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,386 @@
...
+                    is uk.akane.libphonograph.dynamicitem.RecentlyAdded -> "android.resource://${context.packageName}/${R.drawable.ic_default_cover_playlist_recently}".toUri()
+                    is uk.akane.libphonograph.dynamicitem.Favorite -> "android.resource://${context.packageName}/${R.drawable.ic_default_cover_playlist_favorite}".toUri()
+                    else -> null
+                }
+                val id = when (item) {
+                    is uk.akane.libphonograph.dynamicitem.RecentlyAdded -> "playlist_recently_added"
+                    is uk.akane.libphonograph.dynamicitem.Favorite -> "playlist_favorite"
+                    else -> "playlist_${item.id}"
+                }
+                createFolderItem(id, title, MediaMetadata.MEDIA_TYPE_FOLDER_MIXED, artworkUri = icon, isPlayable = item.songList.isNotEmpty(), isBrowsable = false)
```
nift4: MEDIA_TYPE_PLAYLIST

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,386 @@
...
+            }
+            is Genre -> createFolderItem(
+                "genre_${item.id}", item.title ?: context.getString(R.string.unknown_genre), MediaMetadata.MEDIA_TYPE_FOLDER_MIXED,
```
nift4: MEDIA_TYPE_GENRE

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,386 @@
...
+                subtitle = context.resources.getQuantityString(R.plurals.songs, item.songList.size, item.songList.size),
+                isPlayable = item.songList.isNotEmpty(), isBrowsable = false, artworkUri = null
+            )
+            is Date -> createFolderItem(
+                "date_${item.id}", item.title ?: context.getString(R.string.unknown_year), MediaMetadata.MEDIA_TYPE_FOLDER_MIXED,
```
nift4: MEDIA_TYPE_YEAR

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
nift4: oh I didn't see the different parameters, but there's still some duplication (why not keep the Sync version, remove the Sync suffix from the name, remove the other searchForMediaItem, inline searchMediaItems, and where the current searchForMediaItem is used, you can use `searchForMediaItem(item.requestMetadata.searchQuery?.trim() ?: "")`

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,386 @@
...
+        val list = app.reader.songListFlow.first()
+        // TODO support focus and sub queries (see MainActivity)
+        return if (text == "") sortList(list, LibraryAdapterTypes.SEARCH, Sorter(SongAdapter.MediaItemHelper, null)) else searchMediaItems(text, list)
+    }
+
+    fun addMediaItems(
+        mediaItems: List<MediaItem>
+    ): ListenableFuture<ExpandedMediaItems> {
+        val completion = SettableFuture.create<ExpandedMediaItems>()
+        scope.launch(Dispatchers.Default) {
```
nift4: I recently learnt there's a nice helper for this pattern: (also applies to the other places where this is used)
https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-guava/kotlinx.coroutines.guava/future.html

(other parts of the codebase don't use it yet, but it's nice to do so in new code)

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,386 @@
...
+                    if (expanded.isNotEmpty()) {
+                        resultList.addAll(expanded)
+                    } else if (item.mediaId != MediaItem.DEFAULT_MEDIA_ID) {
+                        val fullSongList = app.reader.songListFlow.first()
+                        val sortedFull = sortList(fullSongList, LibraryAdapterTypes.SONG, Sorter(SongAdapter.MediaItemHelper, null))
+                        val idx = sortedFull.indexOfFirst { it.mediaId == item.mediaId }
+                        if (idx >= 0 && startingIndex == null) {
+                            startingIndex = resultList.size + idx
+                        }
+                        resultList.addAll(sortedFull)
```
nift4: What is this exactly used for / why did the `gramophoneApplication.reader.songListFlow.first()
                            .filter { m -> m.mediaId == it.mediaId }` not suffice? can you add some context? (I think it might conflict with expectations of phone code)

app/src/main/res/values/strings.xml
```diff
@@ -1194,6 +1194,7 @@
     <string name="actions_query_shuffle_specific">shuffle $item_name</string>
     <!-- Tooltip for heart button -->
     <string name="favorite">Favorite</string>
+    <string name="unfavorite">Unfavorite</string>
```
nift4, line 1198: Please add a comment for translators to understand where this is used, similar to the other strings

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
```diff
@@ -0,0 +1,386 @@
...
+            }
+            else -> Triple(emptyList(), -1, false)
+        }
+        if (adapterType == -1) return emptyList()
+        val sorter = Sorter(SongAdapter.MediaItemHelper, null, if (naturalOrder) Sorter.Type.NaturalOrder else null)
+        return sortList(songs, adapterType, sorter)
+    }
+
+    private fun <T> sortList(list: List<T>, adapterType: Int, sorter: Sorter<T>): List<T> {
+        val prefs = PreferenceManager.getDefaultSharedPreferences(context)
```
nift4: Please pass the one obtained in service onCreate in as constructor parameter backed by a field instead of getting new one over and over again

nift4 (changes requested):
Looking great already, I mostly arrived at only style nitpicks being left to complain about :P I think that's all I can think of though. Just one question I have, see below, where I worry about some logic conflict.

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
banksio: It doesn't say on that page that they are AAOS only, but I tested this afternoon and they don't show up.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
banksio: No problem, I was wondering if you might say that. I've moved it.

app/src/main/java/org/akanework/gramophone/logic/LibraryTreeLoader.kt
banksio: If you're talking about loading the whole library in the queue, it's to prevent a situation where the user taps on a song, and the app plays that song, and then stops. However, your syntax for filtering is a good point and can be used here instead.

nift4: Can you gate this new logic behind requestMetadata being non-null, and if it's null, use the previous logic instead? I think that should work for both Auto and phone code. (Auto or media3's MediaSessionLegacyStub specifically will set request metadata, while phone code won't)

nift4:
Noticed an issue: if a song/artist/etc has no cover it displays a triangle warning icon

nift4:
> Noticed an issue: if a song/artist/etc has no cover it displays a triangle warning icon

I think I have an easy idea to fix it though so you can leave that issue in, I'll try something during merge.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
banksio: I've tried this, and it doesn't work.

In Media3, Android Auto acts as a legacy MediaControllerCompat client and translates modern commands to the old legacy architecture via MediaSessionLegacyStub.java. It's got a hardcoded check for `sessionCommand.commandCode == SessionCommand.COMMAND_CODE_CUSTOM` though:

```java
    for (int i = 0; i < customLayout.size(); i++) {
      CommandButton commandButton = customLayout.get(i);
      SessionCommand sessionCommand = commandButton.sessionCommand;
      if (sessionCommand != null
          && commandButton.isEnabled
          && sessionCommand.commandCode == SessionCommand.COMMAND_CODE_CUSTOM  // this is bad
          && ...) {
           // Adds as CustomAction for Android Auto
      }
    }
```
So we have to rely on the custom command here.

nift4: Hmm, it should work as we set mediaButtonPreferences and not the legacy custom layout, so it's auto-converted inside media3. see MediaSessionLegacyStub.updateCustomLayoutAndLegacyExtrasForMediaButtonPreferences CommandButton.getCustomLayoutFromMediaButtonPreferences, and specifically convertToPredefinedCustomCommandButton. I'll debug why it doesn't work

nift4:
Are you still working on anything, or are you done?

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
banksio: Ah, my bad...

`COMMAND_CODE_SESSION_SET_RATING` needs a Rating parameter to be considered executable.

```java
  public boolean canExecuteAction() {
    if (sessionCommand != null) {
      switch (sessionCommand.commandCode) {
        // ...
        case COMMAND_CODE_SESSION_SET_RATING:
          return parameter != null;
        default:
          return false;
      }
    } 
    // ...
```

I've added a parameter and pushed that change.

banksio:
I've just fixed the rating command. Done for now. Apologies for jumping the gun on the review request.

nift4:
>  Apologies for jumping the gun on the review request.

No worries.

> I've just fixed the rating command.

Did you push it?

banksio:
Got my branches mixed up, pushed now.

nift4 (approved):

nift4:
Thanks for the PR, and for being responsive to review feedback. I'm sure there will be quite a few users happy thanks to this.

banksio:
> Thanks for the PR, and for being responsive to review feedback. I'm sure there will be quite a few users happy thanks to this.

No worries, thank you for being patient with me and receptive to my changes!

PR #952, opened 2026-06-19
feat: locale-aware queue titles via QueueTitle sealed class

### LINKED ISSUE #943: [FR] Queue title localization
Where string resources are used, can be done with enum type.

https://github.com/FoedusProgramme/Gramophone/pull/865#discussion_r3357493127

mvanhorn:
## Summary

- Introduce a `QueueTitle` sealed class with two variants: `Resource(@StringRes val resId: Int)` for system-generated titles and `Custom(val name: String)` for user-defined names.
- Replace `MultiQueueObject.title: String` with `MultiQueueObject.title: QueueTitle`; update Bundle serialization with backward-compatible fallback for legacy state.
- Update `AdapterFragment.getQueueTitle()`, `GeneralSubFragment`, `SongAdapter`, `AlbumAdapter`, `ViewPagerFragment`, and `GramophonePlaybackService` to produce and consume `QueueTitle` values instead of resolved strings.
- Resolve titles to display strings at render time in `MqListItem` and `MqState.getQueueTitle(context)` via `QueueTitle.resolve(context)`.

## Why this matters

Queue titles like "Songs" or "Albums" were previously resolved once via `getString()` at the moment a song was played, then baked into `MultiQueueObject.title` as a plain `String`. This prevented proper localization on locale change. See #943.

## Testing

- Play a song from each main tab (Songs, Albums, Genres, Dates, Folders, Playlists) and confirm the queue title shown in the multi-queue view matches the current locale's translation of that tab label.
- Rename a user playlist and play from it; the queue title should show the user-defined name, not a translated resource.
- Change device locale while a queue is active; after relaunching the multi-queue UI the system-generated title should reflect the new locale.
- Verify `MultiQueueObject.title` round-trips correctly through Bundle serialization; legacy saved state (plain string) falls back to `QueueTitle.Custom`.
- Confirm the `(+​)` extension queue suffix logic in `QueueBoard.addQueue` still appends correctly.

Fixes #943

### DIFF
--- a/app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
+++ b/app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
@@ -126,6 +126,7 @@ import org.akanework.gramophone.logic.utils.ReplayGainUtil
 import org.akanework.gramophone.logic.utils.SemanticLyrics
 import org.akanework.gramophone.logic.utils.exoplayer.EndedWorkaroundPlayer
 import org.akanework.gramophone.logic.utils.exoplayer.EndedWorkaroundPlayer.Companion.queueWithTitle
+import org.akanework.gramophone.logic.QueueTitle
 import org.akanework.gramophone.logic.utils.exoplayer.GramophoneExtractorsFactory
 import org.akanework.gramophone.logic.utils.exoplayer.GramophoneMediaSourceFactory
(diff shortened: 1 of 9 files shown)


nift4:
Thanks for spending your tokens on our project, sadly we can't take untested/not building pull requests, we need a human who is interested enough in the project to test and look at the code

PR #997, opened 2026-08-10
fix: correctly handle Android Auto favorite command

banksio:
This PR fixes an issue where pressing the favourite button in Android Auto caused the screen to flicker without actually toggling the favourite status or updating the button's icon.

`GramophonePlaybackService` was missing an override for `onMediaMetadataChanged`, so when the song was added/removed from favourites, the custom layout sent to connected controllers (e.g., Android Auto) was never refreshed.

I've modified the playback service to hook onMediaMetadataChanged and trigger `refreshMediaButtonCustomLayout()`.

### DIFF
--- a/app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
+++ b/app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
@@ -1539,6 +1539,10 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
         }
     }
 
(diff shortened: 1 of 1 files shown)


nift4:
ensure you have notifications for gramophone turned on while testing, so that it can post the missing permission error if needed

banksio:
Ready to go. The original description was wrong, the command was being handled just fine, but the custom layout wasn't being refreshed. Confirmed working locally.

nift4:
that makes sense, thanks

PR #1003, opened 2026-08-22
UI: Add features for player controls, lyrics notifications, and widgets

SteveZMTstudios:
## Summary

- **Card widget** — a new responsive home-screen widget family for quick playback control. One provider delivers pill (1×1), circle (2×2 w/ progress ring), card (4×1), medium (4×2) and large (4×4 / wide 5×4) layouts, plus favorite & shuffle actions on wide slots.
- **Notification lyrics** — opt-in setting that surfaces the current synced lyric line in both the expanded media notification and the system media session metadata (lockscreen / BT / Android Auto), with the original title+artist preserved as the subtitle line.
- **Swipe cover to switch track** on the full player (default enabled); gesture recognition keeps vertical gestures & single-tap behavior intact.
- **Optional rotating cookie play/pause button** to match the cover-rotation feel while playing, controlled by a player setting.
- **"Always skip to previous track" behavior** — opt-in flag so pressing Previous jumps straight to the last item instead of seeking back to track start, useful for heads-free and hardware-button use cases.
- Widget favorite action connects through a dedicated playback-service helper so the current item's heart rating and widget state stay in sync across lyric/session updates.
- Widget resources use Material You system colors on Android 12+ with stable theme-token fallbacks for pre-12 and night mode; default + zh-rCN strings added for all new settings and the widget picker label.

## Test

Installed on my phone (android 16):

Screenshot_2026-08-22-19-54-58-26_b783bf344239542886fee7b48fa4b892

and i run it about 3 hours playing music, no crash and any error occured.

Also Windows subsystem of Android:
image

with music3 patch at https://github.com/nift4/media/pull/2, works on Android 6 AVD:

image

And my low-performance platform (Android 8.1 MT6580 1GB RAM):
Screenshot_20260823-010351_Trebuchet

Even my Android 7.1 arm 512MB RAM device MT6572 1.2GHz x2，

Screenshot_20260823-234356

All works fine and stable.

CI Passed:

https://github.com/SteveZMTstudios/Gramophone/actions/runs/32650576403

PR #1005, opened 2026-08-22
feat(lyrics): add notification lyrics support

SteveZMTstudios:
## Summary

Notification lyrics — opt-in setting that surfaces the current synced lyric line in both the expanded media notification and the system media session metadata (lockscreen / BT / Android Auto), with the original title+artist preserved as the subtitle line.

Screenshots, CI tests and more, see #1003

### DIFF
--- a/app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
+++ b/app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
@@ -219,11 +219,6 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
     private lateinit var prefs: SharedPreferences
     private var lastSentHighlightedLyric: String? = null
     private var lastSentNotificationLyric: String? = null
-    private var isPausedFallback = false
-    private val pauseFallbackRunnable = Runnable {
-        isPausedFallback = true
-        sendLyricNow(false)
-    }
     private lateinit var afFormatTracker: AfFormatTracker
     private lateinit var rgAp: ReplayGainAudioProcessor
     private var rgMode = 0 // 0 = disabled, 1 = track, 2 = album, 3 = smart
@@ -331,11 +326,7 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
         qb = QueueBoard(this)
         setListener(this)
         setMediaNotificationProvider(
-            MeiZuLyricsMediaNotificationProvider(
-                this,
-                tickerProvider = { lastSentHighlightedLyric },
-                notificationLyricProvider = { lastSentNotificationLyric }
-            )
+            MeiZuLyricsMediaNotificationProvider(this) { lastSentHighlightedLyric }
         )
         setForegroundServiceTimeoutMs(120000)
         setShowNotificationForEmptyPlayer(SHOW_NOTIFICATION_FOR_EMPTY_PLAYER_AFTER_STOP_OR_ERROR)
@@ -828,7 +819,6 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
         scope.cancel()
         endedWorkaroundPlayer!!.stop()
         handler.removeCallbacks(timer)
-        handler.removeCallbacks(pauseFallbackRunnable)
         handler.removeCallbacks(sendLyrics)
         mediaSession!!.setOptOutOfMediaButtonPlaybackResumption(controller!!.currentTimeline.isEmpty)
         proxy?.let {
@@ -917,7 +907,7 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
         var restart = false
         if (key == null || key == "notification_lyrics" || key == "status_bar_lyrics") {
             scheduleSendingLyrics(false)
-            endedWorkaroundPlayer?.invalidatePlayerState()
+            endedWorkaroundPlayer?.updateLyricNow()
         }
         if (key == null || key == "rg_mode") {
             rgMode = prefs.getStringStrict("rg_mode", "0")!!.toInt()
@@ -1535,11 +1525,6 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
     }
 
     override fun onPlaybackStateChanged(state: Int) {
-        if (state == Player.STATE_ENDED || state == Player.STATE_IDLE) {
-            isPausedFallback = true
-            handler.removeCallbacks(pauseFallbackRunnable)
-            sendLyricNow(false)
-        }
         if (state == Player.STATE_IDLE) {
             var changed = false
             if (afTrackFormat != null) {
@@ -1604,8 +1589,6 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
             //lyrics = null
             //scheduleSendingLyrics(true)
         }
-        isPausedFallback = false
-        handler.removeCallbacks(pauseFallbackRunnable)
         lastSentNotificationLyric = null
         lastSentHighlightedLyric = null
 
@@ -1643,13 +1626,6 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
     }
 
     override fun onIsPlayingChanged(isPlaying: Boolean) {
-        if (!isPlaying) {
-            handler.removeCallbacks(pauseFallbackRunnable)
-            handler.postDelayed(pauseFallbackRunnable, 3000L)
-        } else {
-            handler.removeCallbacks(pauseFallbackRunnable)
-            isPausedFallback = false
-        }
         scheduleSendingLyrics(false)
         lastPlayedManager.save()
     }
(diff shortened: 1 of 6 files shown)


SteveZMTstudios:
[Gramophone-1.1.2.495b22b-release_notification_lyrics.zip](https://github.com/user-attachments/files/31337925/Gramophone-1.1.2.495b22b-release_notification_lyrics.zip)

nift4:
Hi, is this intended for use with bluetooth cars and Android Auto mainly, or also for phone usage?

I am wondering whether it should be restricted to the Bluetooth and AA session, so that the notification in phone UI continues to not show lyrics.

SteveZMTstudios:
> Hi, is this intended for use with bluetooth cars and Android Auto mainly, or also for phone usage?
> 
> I am wondering whether it should be restricted to the Bluetooth and AA session, so that the notification in phone UI continues to not show lyrics.

First of all, embarrassingly, I don't have a car yet, and I don't have a device that can display Bluetooth media names. So I cant promise that 100% works.


However, my WearOS watch displays lyrics correctly, so I'm not sure if both of these are working correctly.

IMG_0115

SteveZMTstudios:
Additionally, this feature is only available after the lyrics notification switch in Settings - Player - Lyrics is manually turned on.

As far as I know, modifying the publicly announced name of the playback media to display lyrics is a fairly common practice, or at least that's how the commercial streaming media I use works.

I added this feature because it works well with the manufacturer's fluid cloud and offers an experience almost identical to commercial software.

Screenshot_2026-08-25-01-05-10-69_b783bf344239542886fee7b48fa4b892

nift4:
Makes sense, then we can keep it for phone too.

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
```diff
@@ -97,9 +98,11 @@ class EndedWorkaroundPlayer(
     }
 
     fun updateLyricNow() {
-        if (context.packageName == "com.tencent.qqmusic") {
-            invalidateState()
-        }
+        invalidateState()
```
nift4: Can we gate it behind notif lyrics preference or qqmusic spoof package name, instead of always doing it?

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -1518,6 +1535,11 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
     }
 
     override fun onPlaybackStateChanged(state: Int) {
+        if (state == Player.STATE_ENDED || state == Player.STATE_IDLE) {
+            isPausedFallback = true
```
nift4: isPausedFallback logic should probably be moved to EndedWorkaroundPlayer.getState(), which will always automatically be called if pause status changes, so there's no need to do any special logic. Simply use `state.playWhenReady` for the judgement. (If you want to keep the 3000ms backoff you can also use postDelayed inside EndedWorkaroundPlayer without issue)

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -1795,10 +1843,16 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
...
-        if (lastSentHighlightedLyric != highlightedLyric) {
+        val notificationLyric = getActiveNotificationLyric()
+        if (lastSentHighlightedLyric != highlightedLyric || lastSentNotificationLyric != notificationLyric) {
+            val notifLyricChanged = lastSentNotificationLyric != notificationLyric
             lastSentHighlightedLyric = highlightedLyric
+            lastSentNotificationLyric = notificationLyric
             handler.post {
                 endedWorkaroundPlayer?.let {
+                    if (notifLyricChanged) {
+                        it.invalidatePlayerState()
```
nift4: I wonder if this is really needed because status bar lyric seems to work without

app/src/main/res/values-zh-rCN/strings.xml
```diff
@@ -155,6 +155,8 @@
     <string name="tab_order_summary">应用将在启动时显示第一个标签页。分隔线后的标签页将被隐藏。</string>
     <string name="settings_status_bar_lyrics_title">状态栏歌词</string>
     <string name="settings_status_bar_lyrics_summary">启用魅族状态栏歌词（仅适用于部分设备）</string>
+    <string name="settings_notification_lyrics_title">通知栏歌词</string>
```
nift4, line 158: Thanks for contributing translations for the new feature! However, the automated script that updates changes would override the change and delete the translation. Translations have to be contributed through Weblate for the workflow to work - and this can only be done after english strings in PR was merged.

nift4 (changes requested):

ghhccghk:
@nift4 If this is merged, Xiaomi Media Island will be able to share real-time lyrics.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
SteveZMTstudios: It is required to edit here, if not, these change won't shown on my wearos or (i guess) other receiving media status thing.

Translated by Google.
Status bar lyrics only update the notification ticker (NotificationCompat.Builder.setTicker) directly via NotificationManager, without modifying MediaMetadata.
Notification lyrics (as well as lockscreen, BT car display, Android Auto, and system media session) require mutating the MediaMetadata (title => lyric, artist => artist - title) returned by EndedWorkaroundPlayer.getState().In Media3's ForwardingSimpleBasePlayer, calling invalidateState() is necessary to notify Media3 that the player metadata has changed so it dispatches onMediaMetadataChanged / onPlaylistMetadataChanged to the MediaSession and system notification. Without it, the media metadata wouldn't update as lyrics advance.

app/src/main/res/values-zh-rCN/strings.xml
SteveZMTstudios: well-- i speak chinese and i add it for testing and self use :)
i will submit to weblate after merged. Thanks for notice!

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
```diff
@@ -97,7 +100,9 @@ class EndedWorkaroundPlayer(
     }
 
     fun updateLyricNow() {
-        if (context.packageName == "com.tencent.qqmusic") {
+        val isNotificationLyricsEnabled = PreferenceManager.getDefaultSharedPreferences(context)
```
nift4: Please use the PreferenceManager created in playback service (pass is as constructor argument) for performance reason

app/src/main/java/org/akanework/gramophone/logic/utils/exoplayer/EndedWorkaroundPlayer.kt
```diff
@@ -119,6 +124,31 @@ class EndedWorkaroundPlayer(
...
+                val origTitle = superState.currentMetadata.title?.toString() ?: ""
+                val origArtist = superState.currentMetadata.artist?.toString() ?: ""
+                val subtitle = if (origArtist.isNotBlank() && origTitle.isNotBlank()) {
+                    "$origArtist - $origTitle"
+                } else {
+                    origArtist.ifBlank { origTitle }
+                }
+                val metadataWithLyric = superState.currentMetadata.buildUpon()
+                    .setTitle(notifLyric)
+                    .setArtist(subtitle)
```
nift4, line 139: Do you know if it works with setDisplayTitle+setSubtitle but without setTitle+setArtist?

Reason I'm asking: apps like Pano Scrobbler might not work when changing title/artist

SteveZMTstudios: [Gramophone-1.1.2.1666f33-release.apk.zip](https://github.com/user-attachments/files/31554445/Gramophone-1.1.2.1666f33-release.apk.zip)
[snapshot.tar.gz](https://github.com/user-attachments/files/31554706/snapshot.tar.gz)



I forked out and added another switch, compcompiled one for my phone, but that seems glichey, and edit notification seems unstable.

SteveZMTstudios: Screenshot_20260828-204436

Android 7, which higher than minsdk, unsupport this.

Android 16, which brought by OEM, will flickers and displays erratically, as if it's unsure which indicator to follow.

https://github.com/nift4/media/blob/5fcf02d05e7e840e12274473420f6268d2cd2925/libraries/session/src/main/java/androidx/media3/session/DefaultMediaNotificationProvider.java#L601-L635

Perhaps we can explain the destructive nature of this function within the response switch itself.

nift4:
Thanks for the PR!

PR #1006, opened 2026-08-22
feat(widget): add adaptive card & circular desktop widgets

SteveZMTstudios:
## Summary
Card widget — a new responsive home-screen widget family for quick playback control. One provider delivers pill (1×1), circle (2×2 w/ progress ring), card (4×1), medium (4×2) and large (4×4 / wide 5×4) layouts, plus favorite & shuffle actions on wide slots.
Widget favorite action connects through a dedicated playback-service helper so the current item's heart rating and widget state stay in sync across lyric/session updates.
Widget resources use Material You system colors on Android 12+ with stable theme-token fallbacks for pre-12 and night mode; default + zh-rCN strings added for all new settings and the widget picker label.

Screenshots, CI tests and more, see https://github.com/FoedusProgramme/Gramophone/pull/1003


SteveZMTstudios:
[Gramophone-1.1.2.2371d0c-release_desktop_widgets.zip](https://github.com/user-attachments/files/31338164/Gramophone-1.1.2.2371d0c-release_desktop_widgets.zip)

SteveZMTstudios:
[Gramophone-1.1.2.1231f2f-release_desktop_widgets.zip](https://github.com/user-attachments/files/31351545/Gramophone-1.1.2.1231f2f-release_desktop_widgets.zip)

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -1785,10 +1836,13 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
     }
 
     private fun sendLyricNow(new: Boolean) {
-        if (new)
+        if (new) {
             LyricWidgetProvider.update(this)
-        else
+            CardWidgetProvider.update(this)
```
nift4: Why are you updating card widget in sendLyricNow()? Feels like a strange choice thematically. Maybe onMediaMetadataChanged ?

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -804,6 +805,55 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
...
+            }
+            withContext(Dispatchers.Main) {
+                val updatedItem = item.buildUpon().setMediaMetadata(
+                    item.mediaMetadata.buildUpon()
+                        .setUserRating(HeartRating(newIsHeart))
+                        .build()
+                ).build()
+                val currentIndex = endedWorkaroundPlayer?.currentMediaItemIndex ?: 0
+                endedWorkaroundPlayer?.replaceMediaItem(currentIndex, updatedItem)
+                CardWidgetProvider.update(this@GramophonePlaybackService)
```
nift4: This should be done through listener, not in function that changes state. Otherwise widget may not update then favourite state is changed through other ways

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -804,6 +805,55 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
...
+                }
+            }
+            withContext(Dispatchers.Main) {
+                val updatedItem = item.buildUpon().setMediaMetadata(
+                    item.mediaMetadata.buildUpon()
+                        .setUserRating(HeartRating(newIsHeart))
+                        .build()
+                ).build()
+                val currentIndex = endedWorkaroundPlayer?.currentMediaItemIndex ?: 0
+                endedWorkaroundPlayer?.replaceMediaItem(currentIndex, updatedItem)
```
nift4: Can't all of this code be replaced with call to `onSetRating()` which already does all that?

app/src/main/java/org/akanework/gramophone/ui/CardWidgetProvider.kt
```diff
@@ -0,0 +1,243 @@
...
+    private fun loadArtworkAndRefresh(
+        context: Context,
+        appWidgetManager: AppWidgetManager,
+        appWidgetId: Int,
+        state: CardWidgetPlaybackState,
+        actions: CardWidgetActions
+    ) {
+        val uri = state.artworkUri ?: return
+        CoroutineScope(Dispatchers.IO).launch {
+            val request = ImageRequest.Builder(context)
```
nift4, line 161: I'm wondering if there's a more optimal way to do it, did you test letting image view load the cover itself? (I guess a placeholder can be achieved by stacking two imageview in a framelayout, but IDK if that is a great idea)

nift4 (commented):
Note I didn't have time to go through all of the code yet, this is only partial comments

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
SteveZMTstudios: The main reason is that one of the responsive layout components (the circular layout `card_widget_circle`) contains a dynamic circular progress bar to visualize the playback progress, achieving the effect of a Material Design 3 promotional video. To ensure the progress bar advances smoothly during playback, this component needs to update periodically (approximately once per second).

I utilized the `scheduleSendingLyrics()` function because the service already has an activity handler loop associated with `isPlaying` and the playback position, thus avoiding the creation of redundant timer loops and additional wake-up operations.

However, you are correct; I will work on resolving this issue to avoid the hybrid approach.

nift4: Widget should optimally not depend on it, do you think it's possible to use a ProgressBar with optimized drawable, because a ProgressBar can auto advance timing.

app/src/main/java/org/akanework/gramophone/ui/CardWidgetProvider.kt
SteveZMTstudios: My initial attempt involved handling album art loading directly with RemoteViews/ImageViews, but I encountered noticeable UI flickering issues: Whenever controls were updated (e.g., scrolling lyrics from a lyrics control, progress updates, or play/pause switching), reapplying RemoteViews caused the launcher's `ImageView` to reload/redraw the URI from scratch, resulting in very noticeable and distracting flickering of the album art.

Stacking two `ImageViews` within a `FrameLayout` was completely unnecessary, not only adding unnecessary layout overhead but also failing to prevent the top album art view from flickering during subsequent RemoteViews updates, as they refreshed together.

By loading the album art once in the background using Coil and caching it in memory (`cachedArtworkBitmap`), subsequent control updates (progress bar jumps, playback status changes) could immediately reuse the existing Bitmap, completely eliminating flickering and making updates visually smoother.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
SteveZMTstudios: Android's RemoteViews seems not appear to support automatically advancing progress bars (the only automatically advancing widget in RemoteViews is a text-based timer). The automatically advancing progress bar in system media notifications is handled internally by SystemUI via PlaybackState timestamps, a feature that third-party AppWidgets cannot utilize.

SteveZMTstudios: Or maybe we should remove the progress bar? My head is spinning.

The [promotional video](https://youtu.be/UHQPdP8qgrk?t=17&si=uzaK0gg55rxWB79k) didn't deliberately show a progress bar; instead, they rotated the album cover.

nift4: https://developer.android.com/reference/android/widget/RemoteViews

according to this, https://developer.android.com/reference/android/widget/ProgressBar is supported

nift4: Also see https://stackoverflow.com/a/27269329 for possible approach to render circle using ProgressBar

nift4: Sorry, I just remembered automatic animation is only possible in indeterminate mode, I will think about it some more...

SteveZMTstudios: It can't just tick forward itself. It needs events to dynamically change its progress. Therefore, it can't be designed to move a specific number of squares per second and be controlled by events. And its `indeterminate` is an animation similar to a loading circle.

I think I understand what you mean: use a progress bar instead of a bitmap and then combine the `incrementProgressBy(int)` function to move the progress bar?

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -804,6 +805,262 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
...
+                            .build())
+                        .also { completer.addCancellationListener({ it.dispose() }, mainExecutor) }
+                    "coil load for $uri"
+                }
+            }
+
+            override fun supportsMimeType(mimeType: String): Boolean = isBitmapFactorySupportedMimeType(mimeType)
+            override fun loadBitmapFromMetadata(metadata: MediaMetadata): ListenableFuture<Bitmap>? =
+                metadata.artworkUri?.let { loadBitmap(it) }
+        })
```
nift4: These changes don't seem relevant to this PR

SteveZMTstudios: These methods aren't newly added features; previously they were defined inline within `onCreate()` inside `setBitmapLoader(CacheBitmapLoader(object : BitmapLoader { ... }))`. During refactoring, this logic was extracted into the private helper function `createMediaSessionBitmapLoader()` with single-expression syntax.

nift4: Refactoring and adding new features shouldn't be done at the same time because it's confusing if someone looks into the history later. Also, most importantly, this function is never called and the old code is still there... (If you're using AI, please remember to double check its output, thanks)

nift4:
There's still unused methods fyi, and a conflict because your other PR was merged

nift4:
I'm confused, do you perhaps have different code locally? because in this branch there's still unused functions like createErrorNotificationChannel or initCustomCommands

SteveZMTstudios:
I apologize, I (and my computer) have been handling too many things simultaneously lately, and I made a mistake while sorting the code. I have fixed it and will carefully review the subsequent steps.

> do you perhaps have different code locally?

And yes, I always bring up the code collection in the feature/stevezmt branch change, which is #1003, to test their coupling.

app/src/main/java/org/akanework/gramophone/logic/GramophonePlaybackService.kt
```diff
@@ -808,6 +809,87 @@ class GramophonePlaybackService : MediaLibraryService(), MediaSessionService.Lis
         return onSetRating(session, controller, mediaItemId, rating)
     }
 
+    fun toggleCurrentItemFavorite() {
```
nift4, line 812: can you replace this method with calling onSetRating(), it would avoid needing to add second copy of updateFavoritePlaylist and notifyFavoriteError

nift4:
Sorry for the delay, currently occupied with something else, will review soon.

nift4:
Thanks for your patience, and sorry for the delay. I tried the widgets on my phone now.

My thoughts from testing:

Overall, I really like the concept of the different widgets. But, I think each type of widget at least needs one provider (so one does pill/circle and another does the card ones). Otherwise I think users would simply not notice that there is different types.

Also, I was wondering about the favorite and shuffle button. I think I would prefer there to either be both shuffle and repeat, or none of them. So maybe, in card layout, there could be repeat in place of current favorite, and favorite is moved to be at end of the name text view instead?

Also, I noticed a minor issue where the favorite button is a different color (more grey) than the other buttons.

Lastly, after closing the app, the widget became empty - it should preferably show the last played song instead, that will be played after pressing the play button (the latter part with restarting playback is already working, widget just doesn't show it).

app/src/main/java/org/akanework/gramophone/ui/CardWidgetProvider.kt
```diff
@@ -0,0 +1,239 @@
...
+        handleWidgetAction(context, action)
+    }
+
+    private fun handleWidgetAction(context: Context, action: String) {
+        val service = GramophonePlaybackService.instanceForWidgetAndLyricsOnly
+        val player = service?.endedWorkaroundPlayer
+        when (action) {
+            ACTION_PREVIOUS -> {
+                if (player != null) player.seekToPrevious()
+                else sendMediaButtonFallback(context, KeyEvent.KEYCODE_MEDIA_PREVIOUS)
```
nift4, line 62: I think this fallback may not work for previous and next button, because it uses startForegroundService() but only the play keycode will actually start the service, which means there's a crash

```

09-15 13:05:21.428 21710 21710 E AndroidRuntime: FATAL EXCEPTION: main
09-15 13:05:21.428 21710 21710 E AndroidRuntime: Process: org.akanework.gramophone.debug, PID: 21710
09-15 13:05:21.428 21710 21710 E AndroidRuntime: android.app.RemoteServiceException$ForegroundServiceDidNotStartInTimeException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{8ff05c5 u0 org.akanework.gramophone.debug/org.akanework.gramophone.logic.GramophonePlaybackService c:org.akanework.gramophone.debug}
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.app.ActivityThread.generateForegroundServiceDidNotStartInTimeException(ActivityThread.java:2349)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.app.ActivityThread.throwRemoteServiceException(ActivityThread.java:2317)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.app.ActivityThread.-$$Nest$mthrowRemoteServiceException(Unknown Source:0)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2691)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.os.Handler.dispatchMessage(Handler.java:109)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.os.Looper.loopOnce(Looper.java:232)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.os.Looper.loop(Looper.java:317)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.app.ActivityThread.main(ActivityThread.java:8934)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at java.lang.reflect.Method.invoke(Native Method)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:591)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:911)
09-15 13:05:21.428 21710 21710 E AndroidRuntime: Caused by: android.app.StackTrace: Last startServiceCommon() call for this service was made here
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.app.ContextImpl.startServiceCommon(ContextImpl.java:2056)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.app.ContextImpl.startForegroundService(ContextImpl.java:2010)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.content.ContextWrapper.startForegroundService(ContextWrapper.java:860)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.content.ContextWrapper.startForegroundService(ContextWrapper.java:860)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at org.akanework.gramophone.ui.CardWidgetProvider.sendMediaButtonFallback(CardWidgetProvider.kt:205)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at org.akanework.gramophone.ui.CardWidgetProvider.handleWidgetAction(CardWidgetProvider.kt:73)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at org.akanework.gramophone.ui.CardWidgetProvider.onReceive(CardWidgetProvider.kt:53)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.app.ActivityThread.handleReceiver(ActivityThread.java:4930)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.app.ActivityThread.-$$Nest$mhandleReceiver(Unknown Source:0)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2527)
09-15 13:05:21.428 21710 21710 E AndroidRuntime:        ... 7 more
```

app/src/main/java/org/akanework/gramophone/ui/CardWidgetProvider.kt
```diff
@@ -0,0 +1,239 @@
...
+        return PendingIntent.getBroadcast(
+            context,
+            requestCode,
+            intent,
+            PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
+        )
+    }
+
+    private fun sendMediaButtonFallback(context: Context, keyCode: Int) {
+        val serviceIntent = Intent(Intent.ACTION_MEDIA_BUTTON).apply {
```
nift4, line 199: Media3 has a helper for this now https://github.com/androidx/media/commit/1c58fa0d8007a6101f0178cb3427fbdec7d3adc5

SteveZMTstudios:
I will make time to handle this as soon as possible. Thank you for sharing your experience and feedback.

SteveZMTstudios:
Screenshot_2026-09-19-18-34-14-32_b783bf344239542886fee7b48fa4b892

nift4:
Your screenshot looks nice, but can you remove the favorite button from the 2x1 card (only)? It makes the title unreadable.

SteveZMTstudios:
> Your screenshot looks nice, but can you remove the favorite button from the 2x1 card (only)? It makes the title unreadable.

thats 3x1 actually, my screen is 5x6, and 2x1 with cards makes ugly and its feature almost dumplicate with pill widgets.

btw in the previous round of changes, I revisited the color extraction scheme; it now supports extracting colors from the cover art rather than the system theme color, and allows for independent theme configuration for the widget.

It's now looks like.


Screenshot_2026-09-21-22-07-50-45_b783bf344239542886fee7b48fa4b892

Issue #584, opened 2025-07-15
[FR] Add reverse order sorting option to existing sorting choices


xiaocaoooo:
**Is your feature request related to a problem? Please describe.**
Currently, the existing sorting options only support ascending order (e.g., sorting by time from oldest to newest, by numerical value from smallest to largest, or by name in alphabetical order). This brings inconvenience when users need to view content in reverse order (e.g., checking the latest items first, finding the largest values quickly, or viewing names in reverse alphabetical order), as they have to manually adjust or cannot achieve the desired sorting effect directly.  
目前，现有的排序选项仅支持正序排列（例如：按时间从旧到新、按数值从小到大、按名称按字母顺序排列）。当用户需要按倒序查看内容时（例如：优先查看最新项、快速找到最大数值、按字母倒序查看名称等），这种设计会带来不便，因为用户不得不手动调整或无法直接实现期望的排序效果。

**Describe the solution you'd like**
I hope to add a reverse order sorting option based on the existing sorting choices. Specifically:  
- For each existing sorting dimension (such as time, numerical value, name, etc.), add a corresponding reverse order sub-option (e.g., in the sorting dropdown menu, after "Sort by time (oldest first)", add "Sort by time (newest first)").  
- Alternatively, add a universal "Reverse Sort" button next to the existing sorting control. When clicked, it can instantly reverse the current sorting order, switching between ascending and descending order.  
希望在现有排序选项基础上增加倒序排序功能。具体来说：  
- 针对每个现有的排序维度（如时间、数值、名称等），增加对应的倒序子选项（例如：在排序下拉菜单中，在“按时间排序（从旧到新）”后增加“按时间排序（从新到旧）”）。  
- 或者，在现有排序控件旁增加一个通用的“反转排序”按钮，点击后可即时反转当前排序顺序，在正序和倒序间切换。

**Describe alternatives you've considered**
I have considered some alternative solutions, such as manually reversing the sorted results through secondary processing, but this is inefficient and not user-friendly, but it complicates the process and affects the user experience. Therefore, adding a direct reverse order sorting option is the most straightforward and effective solution.  
我曾考虑过一些替代方案，例如通过二次处理手动反转排序结果，但这种方式效率低且不够友好，但流程繁琐，影响用户体验。因此，直接增加倒序排序选项是最直接有效的解决方案。

**Additional context**
[...]

PR #1018, opened 2026-09-01
Add reverse sort order option

manum45:
Adds option to reverse the order type.
This solves #584 
I kept the defaults of whether ascending or descending is used for each sort mode because I think it is helpful, and added a checkbox that causes the inverse of the default to be used (as opposed to adding an ascending <> descending toggle).

Also fixed some bugs in the sorting options.

Note: I made these changes with the help of an LLM, but I checked all of them and adapted where I saw fit.


app/src/main/java/org/akanework/gramophone/ui/adapters/DetailedFolderAdapter.kt
```diff
@@ -261,6 +281,7 @@ class DetailedFolderAdapter(
 
     override fun sort(type: Sorter.Type) {
         sortType.value = type
+        songAdapter.sort(type)
```
nift4: Why change the song list's sort setting when changing the folder sort setting?

app/src/main/java/org/akanework/gramophone/ui/adapters/Sorter.kt
```diff
@@ -136,17 +137,28 @@ class Sorter<T>(
...
 
     fun getSupportedTypes(): Set<Type> {
         return sortingHelper.typesSupported.let { types ->
+            var res = types
             if (naturalOrderHelper != null || rawOrderExposed == Type.NaturalOrder)
-                types + Type.NaturalOrder
-            else types
+                res = res + Type.NaturalOrder + Type.NaturalOrderDescending
+            if (res.contains(Type.ByDiscAndTrack))
+                res = res + Type.ByDiscAndTrackDescending
```
nift4: why special case here?

app/src/main/java/org/akanework/gramophone/ui/adapters/BaseDecorAdapter.kt
```diff
@@ -107,18 +107,18 @@ open class BaseDecorAdapter<T : AdapterFragment.BaseInterface<*>>(
...
-                            buttonMap.entries
-                                .first { it.value == adapter.sortType.value }.key
-                        ).isChecked = true
-                    }
 
-                    else -> throw IllegalStateException("Invalid sortType ${adapter.sortType.value.name}")
-                }
+            val currentSort = adapter.sortType.value
+            val activeEntry = buttonMap.entries.find { it.value == currentSort || Sorter.Type.inverse(it.value) == currentSort }
+            val defaultSort = activeEntry?.value ?: Sorter.Type.None
```
nift4: if you move this assingment below the null-check, you can get rid of the ?: None which is never used

app/src/main/java/org/akanework/gramophone/ui/adapters/BaseDecorAdapter.kt
```diff
@@ -131,16 +131,27 @@ open class BaseDecorAdapter<T : AdapterFragment.BaseInterface<*>>(
...
+                reverseItem.isChecked = currentSort != defaultSort && currentSort != Sorter.Type.None
+            }
+
             popupMenu.setOnMenuItemClickListener { menuItem ->
                 when (menuItem.itemId) {
                     in buttonMap.keys -> {
                         if (!menuItem.isChecked) {
-                            adapter.sort(buttonMap[menuItem.itemId]!!)
+                            val baseType = buttonMap[menuItem.itemId]!!
+                            val targetType = if (reverseItem.isChecked) Sorter.Type.inverse(baseType) ?: baseType else baseType
```
nift4: I'm wondering, wouldn't it be better UX to reset "reverse order" to false when user chooses a different sort type?

Perhaps saving a per-sort option cache to disk, so that following UX is achieved (example of main song list):
1. User starts with sort option "Name", not reversed -> A-Z
2. User enables reverse mode for Z-A order
3. User selects Modified date sort option instead, because user didn't use this option before, reverse is automatically turned off
4. Later, User selects name again, and the reverse state is restored, Z-A order is loaded automatically

nift4 (commented):
Thanks for the PR, just another of these million things you never get around to. I left some comments

app/src/main/java/org/akanework/gramophone/ui/adapters/DetailedFolderAdapter.kt
manum45: I thought both should be changed with the same button. I missed that there's a separate button for the song list. Will undo this change.

app/src/main/java/org/akanework/gramophone/ui/adapters/Sorter.kt
manum45: right that's now redundant with the calculation below. will remove

app/src/main/java/org/akanework/gramophone/ui/adapters/BaseDecorAdapter.kt
manum45: cleaned this up a little

app/src/main/java/org/akanework/gramophone/ui/adapters/BaseDecorAdapter.kt
manum45: I agree this would be the best behavior. But for now I'd like to keep the logic simple, especially since I don't have any experience with the preferences system.
I will implement the reset of the reverse order toggle if sorting type is changed.
Would that be fine for you?

nift4: Ok, merging, I'll do the preference part later. Thanks for the PR! :)

manum45:
Thanks :) should the issue be closed or do you want to keep it open for the preferences?

nift4:
Oh yeah I kinda expected it to auto-close, thanks for the reminder

